PR #87, opened 2015-01-01
Fix keep all tagged logic

SeeSpotRun:
Finally did some proper testing of -km options (fixed plenty of bugs in the process although no dangerous ones)


sahib:
For future pull requests: please use ``sahib:develop`` as target. ``sahib:master`` is where (mostly) stable changes are committed too. 

Also, those changes seem to break some tests in test_merge_directories:
https://travis-ci.org/sahib/rmlint/jobs/45588133

Will merge locally into develop and check what's wrong if there's some spare time today.

sahib:
Should be hopefully fixed in latest develop branch. Your commits are included there too. 

Note: -S a does case insensitive sort now. Since this is probably more useful.

PR #101, opened 2015-01-24
Develop

SeeSpotRun:
Seems to be more or less working; still a few TODO's.


sahib:
I reverted this for now since it broke quite a few things. Fixing one lead to breaking something else for me. For example hardlinks and symlinks lead to an infinite loop after finding everything. Im not sure how the file count semantics are supposed to work here and currently have not the time to check.

PR #102, opened 2015-01-25
Updated the way environment is forwarded to scons.

k0ral:
- Do not raise an error if an environment variable isn't defined (notably: `TERM` should be optional).
- Forward the `PKG_CONFIG_PATH` variable, used by `pkg-config`.

Rationale: I'm packaging `rmlint` for [nix](https://github.com/NixOS/nixpkgs).


sahib:
Thanks for the hint. Merged.

PR #123, opened 2015-05-11
Respect $PKG_CONFIG

Somasis:
Build systems like autoconf and friends respect `$PKG_CONFIG` being set to a different binary path other than just 'pkg-config', which is useful when cross-compiling; we should respect that.


sahib:
Thanks! Merged.

PR #155, opened 2015-10-09
Fix/worth waiting

SeeSpotRun:
Should fix the failure to abort gracefully.  Problem was that https://github.com/SeeSpotRun/rmlint/blob/7425e9f9b4813524a41ee2b43ea7e4ef1ca88e68/lib/shredder.c#L724 is a dumb test given https://github.com/SeeSpotRun/rmlint/blob/7425e9f9b4813524a41ee2b43ea7e4ef1ca88e68/lib/file.h#L174

### DIFF
--- a/lib/cfg.h
+++ b/lib/cfg.h
@@ -118,6 +118,7 @@ typedef struct RmCfg {
     gboolean shred_always_wait;
     gboolean shred_never_wait;
     gboolean fake_pathindex_as_disk;
+    gboolean fake_abort;
 
     /* If true, files are hold back to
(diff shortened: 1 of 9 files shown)


sahib:
Sorry, was away a few days because a construction site managed to cut my DSL cable.

Merged this, abort seems to work fine now (although it takes a little more time than just aborting mds hard, which was fine too imho in case of an SIGINT).

SeeSpotRun:
Yes the extra time is to finish processing data that has already been read, and properly dispose of any unread files and shred groups.

PR #157, opened 2015-10-14
scons: add fts support on musl

thypon:
This patch allows rmlint to compile on musl based linux distribution like voidlinux/alpine/... .

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -302,6 +302,19 @@ def check_linux_limits(context):
     context.Result(rc)
     return rc
 
+def check_musl_fts(context):
+    rc = 1
+    if tests.CheckHeader(context, 'fts.h'):
+        rc = 0
+
+    if tests.CheckLib(context, ['fts']):
+        rc = 0
+
+    conf.env['HAVE_FTS_H'] = rc
+    context.did_show_result = True
+    context.Result(rc)
+    return rc
+
 
 def create_uninstall_target(env, path):
     env.Command("uninstall-" + path, path, [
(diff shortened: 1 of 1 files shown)


sahib:
I'm a bit confused by this commit. Where exactly is the support for musl coming in? 
You're checking for ``fts.h``, but that will never show up in ``config.h`` (see ``lib/SConscript``).

thypon:
Should be fixed.

sahib:
I'm sorry, I still see mostly the same commit as before (65d399f vs. e3953d2). 

To re-iterate: My problem with this commit that I can't see where this actually changes something in the way rmlint is build. I'm already happy with an explanation if I missed something.
Did you test it?

thypon:
What it changes is the linking phase. Musl hasn't an internal fts implementation and needs to link to an external implementation like https://github.com/pullmoll/musl-fts. Another solution could be including fts.{c,h} from musl-fts directly inside rmlint (they are NetBSD fts implementation adapted to run on linux mainly, they works also on glibc).

Btw this patch is in production on voidlinux since 2.2.0. https://github.com/voidlinux/void-packages/tree/master/srcpkgs/rmlint

sahib:
Thanks for the link to the musl-libc, I replaced the system ``fts.h``  with the ``fts.[ch]`` from your link.
Apparently all tests ran through and there was no speed difference. 

This would solve some portability issues we had for more exotic systems, 
therefore I would prefer this solution. 

The license appears to be BSD(-like)?

thypon:
Yep it is.

I have done a similar thing for ulatencyd: https://github.com/thypon/ulatencyd/commit/d8035c6346c4246ab1ae755c910cff029ee34997 .

pullmol's musl-fts implementation is derived from that.

sahib:
I merged my ``feature/musl-fts`` with ``develop``. It will be included in the next release 2.4.0.
Thanks for the help.

thypon:
Thank you for the effort and your nice project.

PR #163, opened 2015-11-14
Cleanup/checksum buffer mem

SeeSpotRun:
A range of tidy-ups and rationalisations.  Only user impact should be removal of --max-paranoid-mem option (https://github.com/sahib/rmlint/commit/73d4ce3940979aed997d006ba0c7bfa6d85bfe06); --limit-mem is now sufficient.


sahib:
Cool: ```+240 −337```.

(Re-added``uniques.c`` which got lost somehow, causing travis to complain)

PR #184, opened 2016-05-31
The gui breaks when a blank optical disc is in the drive as it doesn't have a path

jmdana:
File "/usr/lib/python3.5/site-packages/shredder/views/locations.py", line 360, in add_entry
    path = path.strip()
AttributeError: 'NoneType' object has no attribute 'strip'

This patch fixes the bug by skipping volumes without path.


sahib:
Thanks for fixing this; Likely wouldn't ever have noticed due to the lack of a disc reader in my computer.

PR #188, opened 2016-09-12
Default prefix to /usr

BrunoVernay:
to simply work "out of the box".
It is not complicated, to change the prefix, but it should work by default.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -429,7 +429,7 @@ link_shared_library_message = '%sLinking Shared Library %s==> %s$TARGET%s' % \
 ###########################################################################
 
 AddOption(
-    '--prefix', default='/usr/local',
+    '--prefix', default='/usr',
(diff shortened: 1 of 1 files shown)


sahib:
Sorry, I won't merge this.

I understand the reasoning, but [by convetion](http://www.linuxfromscratch.org/blfs/view/svn/introduction/position.html) ``/usr/local`` should contain self-compiled/user-installed software, while ``/usr`` contains software packaged by the distributions. I'm aware this is not a hard rule followed by everyone, but it kinda makes sense to me. Also, ``/usr/local/bin`` is usually in the ``PATH``, therefore rendering this to a *"cosmetic"* problem.

Thanks for your effort though.

BrunoVernay:
No problem. It is fine.  (Wasn't that much effort.)

PR #189, opened 2016-09-12
Indicate how to clean

BrunoVernay:
Cleaning is not obvious. I had to "man scons" to find how to clean.
Usually "make clean" does the job, but scons is not well known.


sahib:
Thanks, merged.

PR #204, opened 2017-01-15
Add keywords to desktop file.

e7appew:
Lintian requires that desktop files shipped with Debian packages have a Keywords entry. This is what I have added for now. Feel free to change it to ones you think are more appropriate.

Cheers.


sahib:
Looks good to me.

Issue #207, opened 2017-02-16
Does not build on GNU Hurd: lib/fts/fts.c:135:48: error: 'MAXPATHLEN' undeclared


xtaran:
[rmlint fails to build from source on Debian GNU/Hurd](https://buildd.debian.org/status/fetch.php?pkg=rmlint&arch=hurd-i386&ver=2.4.6-1&stamp=1487201702&raw=0) due to using `MAXPATHLEN` (which does not exist on GNU Hurd) unconditionally:

```
gcc -o lib/fts/fts.o -c -Wall -W -Wextra -Winit-self -Wstrict-aliasing -Wmissing-include-dirs -Wuninitialized -Wstrict-prototypes -g -O2 -fdebug-prefix-map=/<<PKGBUILDDIR>>=. -specs=/usr/share/dpkg/pie-compile.specs -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -std=c11 -pipe -D_GNU_SOURCE -fPIC -Os -pthread -I/usr/include/blkid -I/usr/include/uuid -I/usr/include/json-glib-1.0 -I/usr/include/gio-unix-2.0 -I/usr/include/glib-2.0 -I/usr/lib/i386-gnu/glib-2.0/include lib/fts/fts.c
gcc: note: pie specs /usr/share/dpkg/pie-compile.specs ignored when pie is not enabled
In file included from lib/fts/fts.c:40:0:
lib/fts/fts.c: In function 'fts_open':
lib/fts/fts.c:135:48: error: 'MAXPATHLEN' undeclared (first use in this function)
     if(fts_palloc(sp, MAX(fts_maxarglen(argv), MAXPATHLEN)))
                                                ^
lib/fts/fts.c:135:48: note: each undeclared identifier is reported only once for each function it appears in
scons: *** [lib/fts/fts.o] Error 1
scons: building terminated because of errors.
```

This seems to be a widespread and well-known issue, as it's [mentioned  in the GNU Hurd Porting Guidelines](https://www.gnu.org/software/hurd/hurd/porting/guidelines.html#PATH_MAX_tt_MAX_PATH_tt_MAXPATHL).

If someone provides a patch, I can test on a Debian GNU/Hurd porterbox if it fixes the issue.

sahib:
Okay, I see. I wonder how many GNU/Hurd users of ``rmlint`` are out there. :smile:
Will fix that on the weekend, should be easily doable.

[...]

PR #208, opened 2017-02-16
Fix FTBFS for hurd-i386.

e7appew:
Hurd does not have path length restrictions, so doesn't define
MAXPATHLEN in <sys/param.h>. Defining it as 4096 when not defined
is a sufficient workaround, since the code is able to adapt when
paths are actually longer than this value.


e7appew:
Sorry, I didn't realise there is also a problem with PATH_MAX. Closing pull request as this would be covered by #207.

PR #210, opened 2017-02-20
Prevent app window from getting stuck in fullscreen mode.

e7appew:
I'm not sure if this is the best solution to this problem, but in Xfce you can end up stuck in fullscreen mode with no way out but to close (using keyboard shortcuts) and restart the application. This is because the title bar is not displayed in fullscreen mode, and without it you have no access to the window menu which pops up when you right click on the title bar.

I couldn't find any window hints that might help with this problem, so the way I went about it was to toggle fullscreen mode and maximise the window instead whenever the user tries to switch to fullscreen mode. However, if fullscreen mode really does need to be made available, I thought perhaps there could be a shortcut key like F11 to toggle it back.

Thanks.

### DIFF
--- a/gui/shredder/window.py
+++ b/gui/shredder/window.py
@@ -18,6 +18,7 @@ from gettext import gettext
 # External:
 from gi.repository import Gtk
 from gi.repository import Gio
+from gi.repository import Gdk
 
 # Internal:
 import shredder
@@ -260,6 +261,7 @@ class MainWindow(Gtk.ApplicationWindow):
         self.headerbar.pack_end(menu_button)
         self.headerbar.pack_end(search_button)
         self.add(self.main_grid)
+        self.connect('window-state-event', self._on_window_state_event)
 
     def add_header_widget(self, widget, align=Gtk.Align.END):
         """Add a widget to the header, either left or right of the title.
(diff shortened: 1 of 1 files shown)


e7appew:
Requested changes done.

Thanks.

sahib:
Thank you for your changes. I will test your changes in the next few days, once I have access to my normal computer.

sahib:
I only had a quick look at this and while this solves the issue for ``xfce`` it disables the ability to use shredder in fullscreen. This feels pretty hacky to me. The better idea would be probably to show the header bar in fullscreen mode. This would require us to reparent the GtkHeaderBar to the the mainwindow in fullscreen mode. 

Do you have this problem in other GTK3 Apps? If so, would be ``rmlint`` the correct place to report this bug?

(Sorry for the delay, but I'm moving, so this might take a few weeks currently...)

sahib:
I had a quick look again and things look not that good.
We cannot re-parent the headerbar properly once it was set using ``set_titlebar()``,
so I don't see a good way to toggle the headerbar visibility between fullscreen/not fullscreen.

I think this is a more general bug, try for example fullscreening ``gtk3-demo``. Same behaviour there,
which makes me suspect that this is a general behaviour of gtk3-apps with a headerbar.

Closing therefore, since this should be resolved elsewhere.

PR #218, opened 2017-04-05
Tests/salvaged

SeeSpotRun:
This PR salvages some tests I added during my meanderings last year.  I have commented out a couple of these that fail under sahib/develop/rmlint.  There are a couple of TODO's (for treemerge I suspect) as a result.


sahib:
:+1:

PR #220, opened 2017-04-07
Fix/mtime sorting

SeeSpotRun:
This was causing a Heisenbug in nosetests


sahib:
Thanks for finding that. I suspect that broke some travis builds, but I didn't care enough to spend time on it. :+1:

PR #228, opened 2017-06-04
Use #if not #ifdef with HAVE_SYSMACROS_H

ilovezfs:
If scons sets HAVE_SYSMACROS_H to 0, that still counts as being defined.
So use #if instead of #ifdef.

Prevents build failure on macOS.


sahib:
Right, thanks for spotting this.

ilovezfs:
Thanks for the merge!

PR #240, opened 2017-06-26
Improvements to python script

### LINKED ISSUE #239: py formatter: handling bad uid/gid is broken
Thanks for the 2.6.1 release! I'm working on a pull request to fix this issue, and to improve and clean up other things in the generated python script. I'm just filing the issue so it is known that the problem exists and work is being done on it. (I'm not sure if many people are using the python script at the moment.) It will be several days before the pull request is ready, though.

Issue:
Handling of bad uids/gids is currently broken in the python script because it calls `os.chmod()` instead of `os.chown()`.

hungrywolf27:
Closes #239.

### DIFF
--- a/lib/formats/py.py
+++ b/lib/formats/py.py
@@ -170,9 +170,21 @@ def exec_operation(item, original=None, args=None):
         )
 
 
-def main(args, header, data, footer):
+def main(args, data):
     seen_cksums = set()
     last_original_item = None
+
+    # Process header and footer, if present
+    header, footer = [], []
+    if data[0].get('description'):
+        header = data.pop(0)
+    if data[-1].get('total_files'):
+        footer = data.pop(-1)
+    # TODO: Print header and footer data here before asking for confirmation
+
+    if not args.no_ask and not args.dry_run:
+        print('\nPlease hit any key before continuing to shredder your data.', file=sys.stderr)
+        sys.stdin.read(1)
 
     for item in data:
         if item['type'].startswith('duplicate_') and item['is_original']:
@@ -232,12 +244,8 @@ if __name__ == '__main__':
 
 
     try:
-        if not args.no_ask and not args.dry_run:
-            print('\nPlease hit any key before continuing to shredder your data.', file=sys.stderr)
-            sys.stdin.read(1)
-
         for json_doc in json_docus:
-            main(args, json_doc[0], json_doc[1:-1], json_doc[-1])
+            main(args, json_doc)
 
         if args.dry_run:
             print(
--- a/lib/formats/py.py
+++ b/lib/formats/py.py
@@ -155,6 +155,23 @@ def exec_operation(item, original=None, args=None):
             file=sys.stderr
         )
 
+MESSAGES = {
+    'duplicate_dir':    '{c[yellow]}Deleting duplicate directory:',
+    'duplicate_file':   '{c[yellow]}Deleting duplicate:',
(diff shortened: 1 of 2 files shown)


sahib:
Hello again. Thanks for that PR. I will try to review until the end of this week.

lib/formats/py.py
```diff
@@ -78,165 +80,194 @@ def original_check(path, original, be_paranoid=True):
...
+        except ValueError as err:   # File is not valid JSON
+            print('{}: {}'.format(err, doc), file=sys.stderr)
+            sys.exit(-1)
 
     try:
-        if not args.no_ask and not args.dry_run:
-            print('\nPlease hit any key before continuing to shredder your data.', file=sys.stderr)
-            sys.stdin.read(1)
-
+        print('# This is a dry run. Nothing will be modified.')
```
sahib: Maybe add a colored ``#`` for this one too.

sahib: This should only be printed if we're doing a dry_run...

lib/formats/py.py
```diff
@@ -39,14 +39,16 @@
...
-    'red':    "\x1b[31;01m" if USE_COLOR else "",
-    'yellow': "\x1b[33;01m" if USE_COLOR else "",
+    'red':    "\x1b[0;31m" if USE_COLOR else "",
+    'blue':   "\x1b[1;34m" if USE_COLOR else "",
+    'green':  "\x1b[0;32m" if USE_COLOR else "",
+    'yellow': "\x1b[0;33m" if USE_COLOR else "",
     'reset':  "\x1b[0m" if USE_COLOR else "",
-    'green':  "\x1b[32;01m" if USE_COLOR else "",
-    'blue':   "\x1b[34;01m" if USE_COLOR else ""
 }
```
sahib: ``original_check`` needs to be adjusted to work with duplicated directories.

sahib:
I didn't have that much time to check, but I had a quick over it. Overall it looks good, but some places


There also some issues on my desk:

- There is a bug that currently prevents using something like this: ``# rmlint -O py:/tmp.x.py``.
  (No default outputs are created, other than what ``-O`` docs say).
- ``%`` in the python output are duplicated due to the usage of ``fprintf``. This leads to the double ``%%``
  in the progress output.
- Tests need to be adjusted. Especially some test that assert that the script works for both Py2 and Py3.x.
- Code should be run through ``pylint`` to make it a bit more consistent.

Just a note: The original idea was to output an easily modifiable python script that can be adjusted to ones need. So I'm not necessarily for supporting every case and for exactly reproducing the functionality of the ``sh`` output (although I really approve the more familiar output you've added in this PR).

lib/formats/py.py
hungrywolf27: I haven't attempted this since it would require knowing the path to `rmlint`, which isn't included in the json output. As you commented below, it isn't necessary for the python script to support every feature of the `sh` output, at least for now.

lib/formats/py.py
hungrywolf27: Done!

hungrywolf27: Sorry about that, it's fixed!

lib/formats/py.py
sahib: True, but imho this is needed for correctness, not just as feature. But no blocker for this PR, I can implement that myself.

lib/formats/py.py
sahib: :+1:

PR #245, opened 2017-07-19
Reflinks dev

SeeSpotRun:
Some changes related to reflink-capable filesystems.

Bugfixes:
* rm_session_check_kernel_version(): logic error
* test_sh.py: sometimes made reflinks when it wanted to test hardlinks
* rm_offsets_match(): several bugs

Improvements:
* Use FIDEDUPERANGE ioctl for file cloning/deduping to support additional reflink-capable filesystems (xfs, ocfs2)
* add `--is-reflink` command to test if two files are reflinks

API changes:
* deprecate `--btrfs-clone` in favour of `--dedupe` since it now applies to other fs types


SeeSpotRun:
+737 / −275 should be a bit easier to digest than +2807 / −2612

lib/cmdline.c
```diff
@@ -1379,13 +1286,17 @@ bool rm_cmd_parse_args(int argc, char **argv, RmSession *session) {
         {"partial-hidden"           , 0    , EMPTY     , G_OPTION_ARG_CALLBACK  , FUNC(partial_hidden)           , _("Find hidden files in duplicate folders only")                          , NULL}     ,
         {"mtime-window"             , 'Z'  , 0         , G_OPTION_ARG_DOUBLE    , &cfg->mtime_window             , _("Consider duplicates only equal when mtime differs at max. T seconds")  , "T"}      ,
 
+        /* COW filesystem deduplication support */
+        {"dedupe"                   , 0    , 0         , G_OPTION_ARG_NONE      , &cfg->dedupe                   , _("Dedupe matching extents from source to dest (if filesystem supports)") , NULL}     ,
+        {"dedupe-readonly"          , 'r'  , 0         , G_OPTION_ARG_NONE      , &cfg->dedupe_readonly          , _("(--dedupe option) even dedupe read-only snapshots (needs root)")       , NULL}     ,
```
sahib: ``'r'`` is already being used for ``--hidden``.

lib/cmdline.c
```diff
@@ -1435,24 +1346,30 @@ bool rm_cmd_parse_args(int argc, char **argv, RmSession *session) {
         {NULL                     , 0   , HIDDEN           , 0                     , NULL                         , NULL                                                          , NULL}
     };
 
+    const GOptionEntry deprecated_option_entries[] = {
```
sahib: Anything stopping us from deprecating ``--btrfs-clone`` fully?

lib/session.c
```diff
@@ -33,43 +33,58 @@
...
-    int found_minor = session->kernel_version[1];
+bool rm_session_check_kernel_version(int need_major, int need_minor) {
+    static GOnce once = G_ONCE_INIT;
+    g_once (&once, rm_session_read_kernel_version, NULL);
+    int *version = once.retval;
+    int major = version[0];
+    int minor = version[1];
 
-    /* Could not read kernel version: Assume failure on our side. */
-    if(found_major <= 0 && found_minor <= 0) {
```
sahib: Oops, embarassing. :fearful:

lib/session.c
```diff
@@ -160,3 +173,209 @@ bool rm_session_was_aborted() {
...
+            _("dedupe: error %i: failed to open dest file.%s"),
+            errno,
+            cfg->dedupe_readonly ? "" : _("\n\t(if target is a read-only snapshot "
+                                          "then -r option is required)"));
+        rm_sys_close(source_fd);
+        return EXIT_FAILURE;
+    }
+
+    /* fsync's needed to flush extent mapping */
+    fsync(source_fd);
```
sahib: Please do an error check here and log something at least.

lib/session.c
```diff
@@ -160,3 +173,209 @@ bool rm_session_was_aborted() {
...
+    rm_log_error_line(_("rmlint was not compiled with file cloning support."))
+#endif
+
+    return EXIT_FAILURE;
+}
+
+/**
+ * *********** `rmlint --is-reflink` session main ************
+ **/
+int rm_session_is_reflink_main(RmCfg *cfg) {
```
sahib: Should we make ``--equal`` call to ``--is-reflink`` first for big files (since it might be a shortcut)?
Usecase is ``rmlint.sh`` where we expect two files to be reflinks when doing a clone, but nevertheless do a full ``--equal`` on the original check.

tests/test_mains/test_dedupe.py
```diff
@@ -0,0 +1,166 @@
...
+        assert status_code == 0
+
+
+def is_on_reflink_fs(path):
+    parts = psutil.disk_partitions(all=True)
+
+    # iterate up from `path` until mountpoint found
+    p = path
+    while 1:
+        match = next((x for x in parts if x.mountpoint == p), None)
```
sahib: I like your cleverness here, but Python is made to be , therefore I think something like this might be read easier:

```python
def up(path):
    while path:
        yield path
        path = os.path.dirname(path)
        if path == "/":
            break

def is_on_reflink_fs(path):
    for part in psutil.disk_partitions(all=True):
        for up_path in up(path):
            if part.mountpoint == up_path:
                print("{0} is {1} mounted at {2}".format(path, match.fstype, p))
                return match.fstype in REFLINK_CAPABLE_FILESYSTEMS
    print("no mountpoint found for {0}".format(path))
    return False
````
(Not tested)

Also we might want to consider moving things like ``needs_reflink_fs`` and ``assert_exit_code`` to utils.

File: lib/cmdline.c (hunk 1 above)
Comment (SeeSpotRun) [inline lib/cmdline.c]: Yeah but `--dedupe` is a subcommand which is mutually exclusive with the main rmlint.  The `-r` shortcut will clear `cfg->ignore_hidden` and also set `cfg->dedupe_readonly`.  This will work as expected for both `rmlint -r` and `rmlint --dedupe -r`.  It's a bit hacky but it works.
I did start playing around with `GOptionGroup`'s with the idea of formalising the Usage along the lines of:
```
Usage:
rmlint [options]           Search for duplicate files and other lint; see rmlint --help
rmlint dedupe [options]    Dedupe matching extents of two [or more] files; see rmlint --help-dedupe
rmlint gui                 Start rmlint gui app
etc
```
For now I think it works as-is but maybe I'll come back to this later

lib/cmdline.c
SeeSpotRun: I'd envisaged overlapping them for at least one release to give time for existing users and external documentation (https://btrfs.wiki.kernel.org/index.php/Deduplication) to move across to the new syntax

lib/session.c
SeeSpotRun: I'm sure it happens [more often than you think](https://github.com/sahib/rmlint/blob/master/lib/config.h.in#L35-L40) ;-)

lib/session.c
SeeSpotRun: Ok I'll have a look at this.

Although I think `rmlint.sh` should skip paranoia check if using `rmlint --dedupe` since the dedupe ioctl does its own paranoid checks.

tests/test_mains/test_dedupe.py
SeeSpotRun: I think you missed a word in 

> Python is made to be ,

...readable?
Anyway I'll re-write.

sahib: Ah right, sorry. Python is readable, but I'm not. :)

lib/session.c
sahib: Yes, I saw that you changed that earlier... I should really take courses on version checks. ;)

lib/cmdline.c
sahib: Okay, that's fine from my side.

lib/cmdline.c
sahib: Okay, I see. Maybe add a comment/TODO item there. This feels indeed a bit hacky.

SeeSpotRun:
I think that covers all your comments.

You might want to have a closer look at https://github.com/sahib/rmlint/pull/245/commits/20be7258b85fb3fcd5d9cfea414f245e9f205185

lib/session.c
SeeSpotRun: Done via https://github.com/sahib/rmlint/pull/245/commits/9300623317af9d4f6aab7a395892c073bbbe6ad3.  Also provides speedup for hardlinks and path doubles.

lib/cmdline.c
SeeSpotRun: Done: https://github.com/sahib/rmlint/pull/245/commits/929d6df316570f8ca4d6d2bd4752e789ed954af7

sahib:
Perfect! Thanks.

> I think that covers all your comments.

Yep.

PR #247, opened 2017-07-30
A couple of bugfixes, a testcase and some formatting

SeeSpotRun:
In other words, a small tidy-up for #245

### DIFF
--- a/lib/utilities.c
+++ b/lib/utilities.c
@@ -1088,19 +1088,17 @@ static gboolean rm_util_is_path_double(char *path1, char *path2) {
             rm_util_parent_node(path1) == rm_util_parent_node(path2));
 }
 
-
 RmOffsetsMatchCode rm_offsets_match(char *path1, char *path2) {
-
     int fd1 = rm_sys_open(path1, O_RDONLY);
     if(fd1 == -1) {
         rm_log_perrorf("Error opening %s in rm_offsets_match", path1);
         return RM_OFFSETS_ERROR;
     }
 
-#define RM_RETURN(value)     \
-    {                        \
-        rm_sys_close(fd1);   \
-        return (value);      \
+#define RM_RETURN(value)   \
+    {                      \
+        rm_sys_close(fd1); \
+        return (value);    \
     }
 
     RmStat stat1;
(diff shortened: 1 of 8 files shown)


lib/utilities.c
```diff
@@ -1216,8 +1212,8 @@ RmOff rm_offset_get_from_path(_UNUSED const char *path, _UNUSED RmOff file_offse
     return 0;
 }
 
-int rm_offsets_match(char *path1, char *path2) {
-    return (path1 == path2);
+RmOffsetsMatchCode rm_offsets_match(char *path1, char *path2) {
+    return (path1 == path2) ? RM_OFFSETS_MATCH : RM_OFFSETS_NO_DATA;
```
sahib: Why do we compare here by pointer again?

tests/test_mains/test_dedupe.py
```diff
@@ -160,10 +162,66 @@ def test_dedupe_works():
...
             '--is-reflink', path_a, path_b,
             use_default_dir=False,
             with_json=False,
             verbosity=""
         )
+
+# count the number of line in a file which start with patterns[]
+def pattern_count(path, patterns):
+    counts = [0] * len(patterns)
+    f = open(path, 'r')
```
sahib: ``with open`` is your friend. :) (Sorry, too much python lately...)

lib/utilities.c
SeeSpotRun: [embarrassing](https://github.com/sahib/rmlint/commit/cf89f219eb5de0fd74104848a48924d0652f1fdb) legacy of my non-c background

Will fix... and maybe rename `rm_offsets_match` to `rm_link_type(path1, path2)` which will help avoid confusion

tests/test_mains/test_dedupe.py
SeeSpotRun: python wisdom gladly accepted

SeeSpotRun:
Incorporated your feedback thanks.

Also found another [bug](https://github.com/sahib/rmlint/pull/247/commits/5758330840b51557dcaa0aee6ee7c3a0009ba6b1) along the way.  This would have reported files on different filesystems as reflinks, if they had the same physical offsets.

sahib:
Merged, thanks.

Issue #248, opened 2017-08-15
option --no-hardlinked does not delete all duplicates (& manpage suggestions)


Awerick:
I am not sure whether I interpret the `--no-hardlinked` option correctly but it seems, there is a bug:

#### Setup:
```sh
$ echo "text" > file_A
$ ln file_A file_B
$ echo "text" > file_Z
$ ls -l --inode
33106  […]  2  […]  file_A
33106  […]  2  […]  file_B
33107  […]  1  […]  file_Z
```

#### Test `--no-hardlinked`:
```sh
$ rmlint  -S A  --no-hardlinked

# Duplicate(s):
    ls '/tmp/testdir/file_Z'
    rm '/tmp/testdir/file_B'

==> In total 3 files, whereof 1 are duplicates in 1 groups.
==> This equals 0 B of duplicates which could be removed.
```
I expected the file `file_A` to be removed as well – but it isn't. Is this the intended behavior?

### Overview & Suggestion for the man page
Actually it took me some time to understand the intention of the `--hardlinked`/`--no-hardlinked` option.
If I (now) interpret it correctly, the following tables show the intended meaning of the (default) `--hardlinked` option and the `--no-hardlinked` option (with the bug):
_(For both options I included alphabetical ascending vs descending ranking so that the hardlinked file is one time the original and one time the duplicate.)_

#### `--hardlinked`:
_File vs Option_          |   `--hardlinked`<br/> `-S a`:    | `--hardlinked`<br/> `-S A`:
----------------------------:|:-------------------------------------:|:----------------------------------:
**`file_A`** (inode **6**)**:**  | :white_check_mark:       |   :x:  
**`file_B`** (inode **6**)**:**  |   :x:                                 |   :x:  
**`file_Z`** (inode **7**)**:**  |   :x:                                 | :white_check_mark:

#### `--no-hardlinked` (incl. bug):
_File vs Option_          | `--no-hardlinked`<br/> `-S a`: | `--no-hardlinked`<br/> `-S A`:
----------------------------:|:-------------------------------------:|:----------------------------------:
**`file_A`** (inode **6**)**:**  | :white_check_mark:       |   _???_ :x: :zap:
**`file_B`** (inode **6**)**:**  | _also:_ :white_check_mark: :thought_balloon: |   :x:  
**`file_Z`** (inode **7**)**:**  |   :x:                                 | :white_check_mark:

#### Key:
:white_check_mark: : Considered as an original (will _not_ be removed).
[...]

PR #249, opened 2017-08-17
Develop

SeeSpotRun:
Fix a corner case for #248


sahib:
Thanks for fixing my mess. :)

What's up with ``Afferbeck Lauder``?

SeeSpotRun:
"Afferbeck Lauder" is supposedly [Strine](https://en.wikipedia.org/wiki/Strine) for "Alphabetical Order"

Issue #261, opened 2017-11-05
Paranoia settings are a bit confused


SeeSpotRun:
Current hashes for different paranoia levels are:
```
-PP:       xxhash   (64 bit, non-cryptographic)
-P:        bastard  (256 bit, murmur128+city128)
(default): blake2b  (512 bit, cryptographic)
-p:        sha512   (512 bit, cryptographic) if available else sha256
-pp:       bitwise comparison
```

The `-p` option is actually less secure than the default option, due to the default being upgraded to blake2b.
Branch https://github.com/SeeSpotRun/rmlint/tree/paranoia proposes a new sequence for discussion.  This includes a re-write of bastard hash to provide a more secure option than blake2b:
```
-PP:       spooky   (128 bit, non-cryptographic)
-P:        sha256   (256 bit, cryptographic)
(default): blake2b  (512 bit, cryptographic)
-p:        bastard* (1024 bit, cryptographic, blake2b + sha3-512)
-pp:       bitwise comparison
```

sahib:
You're right there. Forgot to change it in that place.
(Although I'd see ``blake2b`` and ``sha512`` somewhat equal - but confusing nevertheless)

> This includes a re-write of bastard hash to provide a more secure option than blake2b:

Truth be told, I'm still no big fan of that bastard hash. From my point of view it neither adds real security, nor performance advantages. Personally I'd even vote to make a single ``-p`` already do the paranoid comparison with no step in between. Would also make the explanation in the docs a bit simpler.

SeeSpotRun:
> Personally I'd even vote to make a single -p already do the bitwise comparison with no step in between

Agreed, was thinking this also; anyone who can't live with the risk of 512 bit hash collisions isn't going to be satisfied by 1024 bits either.

Let's drop `-p` and `-a bastard`.

SeeSpotRun:
Here's a comparison table based on our standard /usr test with warm caches, R5 1600 cpu:
```
                                           BITS
 HASH        TIME   CPU    RAM(kB)  32  64  128 160 256 384 512  ∞
 ----------------------------------------------------------
 spooky32    3.95   313    169       X
 murmur      3.96   311    169               X
 spooky64    3.96   313    169           X
 spooky128   3.96   312    169               X
 xxhash      3.96   299    167           X
 farmhash    3.97   314    170           X
 city128     3.97   331    172               X
~MD5         4.05   331    217               X
 murmur256   4.06   359    180                       X
[...]

PR #265, opened 2017-11-20
Checksum dev

SeeSpotRun:
Sorry another big PR.  Mainly related to #261 but found a few other things along the way.

Summary of changes:

1. Different idiom in checksum.c: conform all digest types to a standard interface; this avoids multiple `switch(digest->type)` spaghetti code and makes adding/removing checksum types easier/cleaner.
2. Drop checksum types bastard, spooky, city & farmhash due to speed and/or efficiency issues.
3. Add metro (fast non-cryptographic) and highway (fast cryptographic) hashes.
4. Change `-p/-P...` hash options.
5. Added tests for impact of data increment length on checksums
6. Sped up `cumulative` digest algorithm & made more robust
7. Docs & changelog updated

All nosetests passing (other than some memory leaks which I don't think are related to this PR)


CHANGELOG.md
```diff
@@ -4,6 +4,38 @@ All notable changes to this project will be documented in this file.
 
 The format follows [keepachangelog.com]. Please stick to it.
 
+## [2.6.2 Toothless Taipan] -- unreleased
```
sahib: Nice name. :smile:
We had the habit that we only give minor releases a new name (that's only a patch release).
I'd be fine to make a minor release though (i.e. increment to 2.7.0).

CHANGELOG.md
```diff
@@ -4,6 +4,38 @@ All notable changes to this project will be documented in this file.
...
+* New checksum types metro and highway
+* Additional unit tests
+* New option --keep-hardlinked
+* --dedupe option can deduplicate twins on any reflick-capable filesystem
+* --dedupe-readonly option can dedupe files on read-only btrfs snapshots
+
+### Changed
+
+* Checksum types for -P... options (see https://github.com/sahib/rmlint/issues/261)
+* Under-the-hood changes for checksum.c
```
sahib: Since the changelog is something that's supposed to be user faces, I'd not mention things that do not have any effect for users. (Same as for additional unit tests)

docs/rmlint.1.rst
```diff
@@ -153,28 +153,31 @@ General Options
 
     Choose the algorithm to use for finding duplicate files. The algorithm can be
     either **paranoid** (byte-by-byte file comparison) or use one of several file hash
-    algorithms to identify duplicates.  The following well-known algorithms are available:
+    algorithms to identify duplicates.  The following cryptographic algorithms are available:
```
sahib: Not sure if it's of any interest to the user if a hash algorithm is suitable for crypotgraphic usage in the context of ``rmlint```. The only property is that the hash function produces well distributed results. 

Sidenote: I'm not entirely sure if ``Highway`` is a "good" cryptographic hash function. I never heard it in that context,
nor do I read a clear statement about it in the project readme. I could be totally wrong here though.

lib/checksum.c
```diff
@@ -23,11 +23,10 @@
...
 
-/* Welcome to hell!
- *
- * This file is 90% boring switch statements with innocent, but insane code
- * squashed between. Modify this file with care and make sure to test all
- * checksums afterwards.
+/* This file is mostly boring interface definitions to conform all of the
+ * difference hash types to a single interface.
+ *  code except for the paranoid digest
+ * optimisations which are pretty insane.
```
sahib: :smile:

lib/checksum.c
```diff
@@ -42,794 +41,858 @@
...
+typedef struct RmDigestInterface {
+    const char *name;           // hash name
+    const guint bits;           // length of the output checksum in bits (if const)
+    RmDigestLenFunc len;        // return length of the output checksum in bytes
+    RmDigestNewFunc new;        // returns new digest->state
+    RmDigestFreeFunc free;      // frees state allocated by new()
+    RmDigestUpdateFunc update;  // hashes data into state
+    RmDigestCopyFunc copy;      // allocates and returns a copy of passed state
+    RmDigestStealFunc steal;    // writes checksum (as binary) to *result
+} RmDigestInterface;
```
sahib: Good work, I like that.

lib/checksum.c
```diff
@@ -42,794 +41,858 @@
...
+#define RM_DIGEST_CUMULATIVE_T guint64
+#define RM_DIGEST_CUMULATIVE_DATA data64
+#define RM_DIGEST_CUMULATIVE_ALIGN 8
+
+#else
+
+#define RM_DIGEST_CUMULATIVE_T guint32
+#define RM_DIGEST_CUMULATIVE_DATA data32
+#define RM_DIGEST_CUMULATIVE_ALIGN 4
+
```
sahib: Did you try to run/compile that in 32bit envs? Otherwise I foresee that we'll get bug reports from those in about ~2 hours after a release. :)

lib/checksum.c
```diff
@@ -42,794 +41,858 @@
...
+        .new = (RmDigestNewFunc)rm_digest_##ALGO##_new,                         \
+        .free = (RmDigestFreeFunc)rm_digest_##ALGO##_free,                      \
+        .update = (RmDigestUpdateFunc)ALGO##_update,                            \
+        .copy = (RmDigestCopyFunc)rm_digest_##ALGO##_copy,                      \
+        .steal = (RmDigestStealFunc)rm_digest_##ALGO##_steal};
+
+CREATE_BLAKE_INTERFACE(blake2b, BLAKE2B);
+CREATE_BLAKE_INTERFACE(blake2bp, BLAKE2B);
+CREATE_BLAKE_INTERFACE(blake2s, BLAKE2S);
+CREATE_BLAKE_INTERFACE(blake2sp, BLAKE2S);
```
sahib: Good use of macros.

lib/checksum.c
```diff
@@ -42,794 +41,858 @@
...
+
+static const RmDigestInterface *rm_digest_get_interface(RmDigestType type) {
+    static const RmDigestInterface *digest_interfaces[] = {
+        [RM_DIGEST_UNKNOWN] = NULL,
+        [RM_DIGEST_MURMUR] = &murmur_interface,
+        [RM_DIGEST_METRO] = &metro_interface,
+        [RM_DIGEST_METRO256] = &metro256_interface,
+#if HAVE_SSE_4_2
+        [RM_DIGEST_METROCRC] = &metrocrc_interface,
+        [RM_DIGEST_METROCRC256] = &metrocrc256_interface,
```
sahib: Similar to the 32 bit comment: Did you try to compile without sse4.2? We had a nasty bug, we're the official arch package was built with sse4.2 support and thus it crashed violently for some (me) that didn't have the instruction set.

lib/checksum.c
```diff
@@ -42,794 +41,858 @@
...
+
+    return code;
+}
+
+const char *rm_digest_type_to_string(RmDigestType type) {
+    const RmDigestInterface *interface = rm_digest_get_interface(type);
+    return interface->name;
+}
+
+/*  TODO: update or remove? */
```
sahib: Remove. Was a nice thought, but since ``rmlint --hash`` will probably never be a widely used tool, it does not make sense. This was the idea behind it (self describing hashes): https://github.com/multiformats/multihash

lib/session.h
```diff
@@ -114,8 +114,7 @@ typedef struct RmSession {
     RmOff offset_fails;
 
     /* Daniels paranoia */
-    RmOff hash_seed1;
-    RmOff hash_seed2;
+    RmOff hash_seed;
```
sahib: Do we even still use that?

lib/shredder.c
```diff
@@ -1070,7 +1065,7 @@ static void rm_shred_file_preprocess(RmFile *file, RmShredGroup **group) {
 
     /* Create an empty checksum for empty files */
     if(file->file_size == 0) {
-        file->digest = rm_digest_new(cfg->checksum_type, 0, 0, 0, NEEDS_SHADOW_HASH(cfg));
+        file->digest = rm_digest_new(cfg->checksum_type, 0);
```
sahib: What's being put into the ``checksum`` field of the json output currently?
I couldn't see from a quick look at the diff if the shadow hash is still being calculated.

tests/test_mains/test_hash.py
```diff
@@ -0,0 +1,63 @@
...
+        command = cmd.format(increment=INCREMENTS[0], algo=algo, path=a)
+        output0 = subprocess.check_output(command.split())
+        for increment in INCREMENTS[1:]:
+            command = cmd.format(increment=increment, algo=algo, path=a)
+            output = subprocess.check_output(command.split())
+            if(output!=output0):
+                assert False, "{} fails streaming test with increment {}".format(algo, increment)
+                break
+
+@with_setup(usual_setup_func, usual_teardown_func)
```
sahib: This looks like a good usecase of ``@parameterized``:

```python
from parameterized import parameterized

@parameterized(["murmur", "metro", ...])
def test_hash_function(name):
    streaming_compliance_check('murmur')
```

sahib:
> Sorry another big PR. Mainly related to #261 but found a few other things along the way.

No problem. I'm (and my lack of time) is more like the problem here, not your work.

I had a look at the PR and looks overall very solid to me. :+1: 
Had a few comments and questions we can still discuss though.

lib/checksum.c
SeeSpotRun: Ok I'll spin up a 32 bit VM and have a play

CHANGELOG.md
SeeSpotRun: Yes there are enough external changes to justify a minor release

docs/rmlint.1.rst
SeeSpotRun: Yes the highway documentation suggests some cryptographic properties but falls short of making any strong claims.  There is a good article at https://arxiv.org/pdf/1612.06257.pdf (you can skip to the conclusions) that is a bit more explicit - I like the "To the best of our (non-expert) ability, we have analyzed the algorithm and
found no weaknesses"

I think all of our hashes have been shown to give well distributed results, so for a given hash length they should all have similar probability of random collisions.  For resistance to malicious collisions I would expect blake and sha3 to be equivalent to each other and better than the others.  But I guess the best anyone can say is "no known attack that can craft a malicious hash collision".

Anyway will update the manpage to avoid unsupported claims.

lib/checksum.c
SeeSpotRun: I must have had a good teacher (thinking formats.h)

lib/checksum.c
SeeSpotRun: That's an interesting dilemma for package-manager-managed releases.  It seems there is a runtime check (`__builtin_cpu_supports ("sse4.2"`) so maybe I'll rewrite that way.

lib/checksum.c
SeeSpotRun: Done.

lib/shredder.c
SeeSpotRun: Yes paranoid now *always* has a shadow hash.

lib/checksum.c
SeeSpotRun: Well that was fun (debian wheezy 32 bit).  Must say I learnt a few things.  Getting close to happy with the results but still have some bugs to fix.

SeeSpotRun:
Ok now compiles and passes nosetests on 32-bit debian wheezy and also on Travis.  Should be ok to merge now.

sahib:
Thanks for the extensive testing!

PR #276, opened 2018-02-23
Shellcheck

SeeSpotRun:
Found this handy utility (https://github.com/koalaman/shellcheck) and have conformed sh.sh/rmlint.sh accordingly


SeeSpotRun:
... was also going to add a nosetest that calls `shellcheck` on the generated `rmlint.sh` but didn't get around to it.

sahib:
Cool, thank you!

PR #288, opened 2018-06-12
Fix percentage in progress bar on rmlint.sh

aaschmid:
... which is broken due to incorrect quoting. My fix does correct the quoting issue introduced in aa9b063827da6903770a1d36debae71fb669d8a6. 

Also remove trailing spaces.

Please don't hesitate to ask any question or give me feedback on this.

### DIFF
--- a/lib/formats/sh.sh
+++ b/lib/formats/sh.sh
@@ -51,7 +51,7 @@ print_progress_prefix() {
         if [ $((PROGRESS_TOTAL)) -gt 0 ]; then
             PROGRESS_PERC=$((PROGRESS_CURR * 100 / PROGRESS_TOTAL))
         fi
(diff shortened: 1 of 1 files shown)


sahib:
Hello @aaschmid,

thanks for the fix. Couldn't test the change yet though, will do later.
Two things:

1) What exactly was broken?
2) The trailing spaces are [sadly needed due to a hack](https://github.com/sahib/rmlint/blob/1a6148c70aa942773a6a9f793266be289f402161/lib/formats/sh.c.in#L460) we're applying to write the correct ``PROGRESS_TOTAL`` to the final script. Please revert that part.

sahib:
Thanks, now I get it. Hard to judge by looking at it from the phone. :smile: 
Approved.

PR #290, opened 2018-06-17
Hardlinks

### LINKED ISSUE #275: MacOS: "cp: illegal option -- -" when trying to replace files with hardlinks
When rmlint is set to replace duplicates with hardlinks, it shows the following error:

[  0%] Hardlinking to original:  /Volumes/Folder/File
cp: illegal option -- -
usage: cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file target_file
       cp [-R [-H | -L | -P]] [-fi | -n] [-apvXc] source_file ... target_directory

My guess: The difference between the BSD (OSX) and GNU (Linux) implementation of the cp command. 

Any ideas?

SeeSpotRun:
Should fix #275.
Also changes handling of duplicate dirs for `cp_hardlink()` and `cp_reflink()` from deletion to symlink.  Arguably neither deletion nor symlink is ideal; we could instead consider banning combinations of options `-D` with `-c sh:hardlink` but I think the symlink option is a reasonable compromise.

### DIFF
--- a/lib/formats/sh.sh
+++ b/lib/formats/sh.sh
@@ -179,12 +179,11 @@ cp_symlink() {
     echo "${COL_YELLOW}Symlinking to original: ${COL_RESET}$1"
     if original_check "$1" "$2"; then
         if [ -z "$DO_DRY_RUN" ]; then
-            touch -mr "$1" "$0"
-            if [ -d "$1" ]; then
-                rm -rf "$1"
-            fi
-            cp --remove-destination --archive --symbolic-link "$2" "$1"
-            touch -mr "$0" "$1"
+            # replace duplicate with symlink
+            rm "$1"
+            ln -s "$2" "$1"
+            # make the symlink's mtime the same as the original
+            touch -mr "$2" -h "$1"
         fi
     fi
 }
--- a/lib/formats/sh.sh
+++ b/lib/formats/sh.sh
@@ -174,6 +174,11 @@ cp_symlink() {
 }
 
 cp_hardlink() {
+    if [ -d "$1" ]; then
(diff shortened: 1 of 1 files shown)


lib/formats/sh.sh
```diff
@@ -159,37 +159,48 @@ original_check() {
...
+            ln -s "$2" "$1"
+            # make the symlink's mtime the same as the original
+            touch -mr "$2" -h "$1"
         fi
     fi
 }
 
-cp_symlink() {
+cp_hardlink() {
+    if [ -d "$1" ]; then
```
sahib: This might not be very obvious to the user. I think the proper way would be to replace every file in the directory with a hardlink... That's quite a bit more complex to implement though.

sahib: After some more thinking about this (and trying to come up with a proper solution) I changed my mind about this a bit. While it's still not very obvious for the user, it's a relatively sane way to handle those edge cases.

lib/formats/sh.sh
```diff
@@ -159,37 +159,48 @@ original_check() {
...
-            touch -mr "$1" "$0"
+            # If it's a directory cp will create a new copy into
+            # the destination path. This would lead to more wasted space...
             if [ -d "$1" ]; then
                 rm -rf "$1"
             fi
-            cp --remove-destination --archive --symbolic-link "$2" "$1"
-            touch -mr "$0" "$1"
+            # replace duplicate with hardlink
+            rm "$1"
```
sahib: Any  reason not to do "rm -rf $1" in both cases (directory and file)?

PR #310, opened 2018-09-24
A quick clean up

mfwitten:
This series fixes some issues that are causing compiler warnings, and it also potentially avoids some run-time bugs; one subtle bug could eventually arise if *all* debug assertions were elided, because its argument is not just a declaration of some condition, but rather an important call with side effects.

Here's a quick summary of the patches:
1. SConstruct: Use `-Wno-cast-function-type' only if GCC's major version >= 8
2. lib/session.c: dedupe: Format strings with `G_GINT64_FORMAT'
3. C: Fix up asserts
4. C: Remove the `rm_assert_gentle*()' macros

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -36,6 +36,28 @@ Export('VERSION_MAJOR VERSION_MINOR VERSION_PATCH VERSION_NAME')
 #                                Utilities                                #
 ###########################################################################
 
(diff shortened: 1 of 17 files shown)


sahib:
Thank you, looks all good to me.

PR #311, opened 2018-09-26
Inline the abort functions

mfwitten:
This series cleans up and inlines the following functions:

* `rm_session_abort()`
* `rm_session_was_aborted()`

The warning messages produced by the latter (including the translations) have also undergone a minor cleanup with regard to punctuation and capitalization.

Also, this series introduces a portable means by which to force the compiler to inline a function.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -40,7 +40,7 @@ def check_gcc_version(context):
     context.Message('Checking for GCC version... ')
 
     try:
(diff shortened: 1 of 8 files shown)


mfwitten:
I've now shoehorned into this pull request a simplified version of `check_gcc_version()` from PR #310; it no longer uses `tail`, but rather just the preprocessor flag `-P`.

sahib:
Thanks. Looks also good to me.

PR #313, opened 2018-09-28
Update and ease support of gettext translations

mfwitten:
This series implements the means by which `gettext` translation files can be updated with one command:

    $ scons gettext

It also uses this new facility to update the `po/*.po[t]` files (however, no new translations are provided).


sahib:
Nice. Thanks for that patch once more.

PR #315, opened 2018-09-29
Cleanup in treemerge

mfwitten:
This series mainly improves `rm_tm_count_files()` by removing an unnecessary parameter, avoiding re-calculation, and introducing [slightly more] graceful error handling. It may be helpful to consult the commit mesage in the following commit of the series:

* lib/treemerge.c: remove unnecessary parameter and recalculation of `path_count' (a5ef34b881d95e41b08ae5a91c92cb0d8cb47517)

### DIFF
--- a/lib/fts/fts.c
+++ b/lib/fts/fts.c
@@ -62,7 +62,7 @@ static FTSENT *fts_build(FTS *, int);
 static void fts_free(FTSENT *);
 static void fts_lfree(FTSENT *);
 static void fts_load(FTS *, FTSENT *);
-static size_t fts_maxarglen(char *const *);
+static size_t fts_maxarglen(const char *const *);
 static size_t fts_pow2(size_t);
 static int fts_palloc(FTS *, size_t);
 static void fts_padjust(FTS *, FTSENT *);
@@ -107,7 +107,7 @@ static int fts_safe_changedir(const FTS *, const FTSENT *, int, const char *);
 #undef FTS_WHITEOUT
 #endif
 
-FTS *fts_open(char *const *argv, int options,
+FTS *fts_open(const char *const *argv, int options,
               int (*compar)(const FTSENT **, const FTSENT **)) {
     FTS *sp;
     FTSENT *p, *root;
@@ -1152,7 +1152,7 @@ static void fts_padjust(FTS *sp, FTSENT *head) {
     }
 }
 
-static size_t fts_maxarglen(char *const *argv) {
+static size_t fts_maxarglen(const char *const *argv) {
     size_t len, max;
 
     _DIAGASSERT(argv != NULL);
--- a/lib/treemerge.c
+++ b/lib/treemerge.c
@@ -178,12 +178,13 @@ int rm_tm_count_art_callback(_UNUSED RmTrie *self, RmNode *node, _UNUSED int lev
     return 0;
 }
 
-static bool rm_tm_count_files(RmTrie *count_tree, GSList *paths, const RmCfg *const cfg) {
+static bool rm_tm_count_files(RmTrie *count_tree, const RmCfg *const cfg) {
     /* put paths into format expected by fts */
 
     g_assert(cfg);
+    guint path_count = cfg->path_count;
+    const GSList *paths = cfg->paths;
 
-    guint path_count = g_slist_length(paths);
     char **path_vec = g_malloc0(sizeof(char *) * (path_count + 1));
     for(guint idx = 0; paths && idx < path_count; idx++, paths = paths->next) {
         path_vec[idx] = ((RmPath *)paths->data)->path;
@@ -260,7 +261,7 @@ static bool rm_tm_count_files(RmTrie *count_tree, GSList *paths, const RmCfg *co
 
     /* Now flag everything as a no-go over the given paths,
      * otherwise we would continue merging till / with fatal consequences,
-     * since / does not have more files as paths[0]
+     * since / does not have more files than path_vec[0]
      */
     for(int i = 0; path_vec[i]; ++i) {
         /* Just call the callback directly */
@@ -521,7 +522,7 @@ RmTreeMerger *rm_tm_new(RmSession *session) {
     rm_trie_init(&self->dir_tree);
     rm_trie_init(&self->count_tree);
 
-    rm_tm_count_files(&self->count_tree, session->cfg->paths, session->cfg);
+    rm_tm_count_files(&self->count_tree, session->cfg);
 
     return self;
 }
--- a/lib/treemerge.c
+++ b/lib/treemerge.c
@@ -178,8 +178,11 @@ int rm_tm_count_art_callback(_UNUSED RmTrie *self, RmNode *node, _UNUSED int lev
     return 0;
 }
 
-static bool rm_tm_count_files(RmTrie *count_tree, GSList *paths, RmSession *session) {
+static bool rm_tm_count_files(RmTrie *count_tree, GSList *paths, const RmCfg *const cfg) {
     /* put paths into format expected by fts */
+
+    g_assert(cfg);
+
     guint path_count = g_slist_length(paths);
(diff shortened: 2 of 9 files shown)


lib/fts/fts.c
```diff
@@ -62,7 +62,8 @@ static FTSENT *fts_build(FTS *, int);
 static void fts_free(FTSENT *);
 static void fts_lfree(FTSENT *);
 static void fts_load(FTS *, FTSENT *);
-static size_t fts_maxarglen(char *const *);
+typedef const char *Path;
```
sahib: This typedef is already in ``fts.h``.

lib/treemerge.c
```diff
@@ -178,20 +178,28 @@ int rm_tm_count_art_callback(_UNUSED RmTrie *self, RmNode *node, _UNUSED int lev
...
-    }
 
-    if(*path_vec == NULL) {
-        rm_log_error("No paths passed to rm_tm_count_files\n");
-        g_free(path_vec);
+    g_assert(cfg);
+    guint path_count = cfg->path_count;
+    g_assert(path_count);
+
+    typedef const char *Path;
```
sahib: Same here.

sahib:
Looks good to me, except minor remarks about seemingly excessive ``typedefs``.

Thanks again.

mfwitten:
I argued with myself over this very question, and ultimately decided the following: Though a `typedef` is frequently used when introducing a new type to the program, it is actually only meant to provide a very weak encapsulation of some *existing* type, a purpose that is useful in 3 cases:

* To obviate the need to write `struct` or `union` when declaring an object. This is probably a more modern usage, especially because C++ doesn't require you to do that; of course, there is a lot of debate on this topic among C aficionados, especially when they are working on a system with constrained memory resources where it is good to be reminded that an object on the stack might be quite large.
* To abstract away an underlying type, so that generic code can be written, and thereby be ported simply by changing that underlying type; it moves the issue from the preprocessor to the C compiler. I suspect this was one of the motivations for creating `typedef`.
* To provide documentation and abstraction of a derived type whose complexity makes it too cumbersome to use C's usual recursive syntax; this is evidenced by the fact that the same `typedef` name can be redeclared multiple times in the same scope (like a forward declaration; the key here is that it is a name *declaration*, not an object or type *definition*). I suspect this was another one of the motivations for creating `typedef`, if not the *main* motivation.

In this case, I decided that I was using the `typedef` in the last way, strictly to convey to the reader a clearer understanding of what is going on; it's a fancy comment that not only allows for simpler declarations, but also allows the reader to find out exactly what a `Path` is supposed to be without having to go grepping through header files ("Oh. It's just a C string!")&mdash;putting it in one centralized location would be about as useful as putting a shared comment in one centralized location; the repetition is the feature, not the mistake.

sahib:
Oh, I did not expect a small essay on the topic. :smile:

You make a point by saying that a ``typedef`` can be used like a comment. 
Still, this results in other questions to ask:

- If the reader is supposed to think *Oh, it's just a* ``const char *Path``, why not directly write  ``const char *Path``? Also, looking the actual type up only needs to happen once. I don't need to be reminded in every function/file where this type is used.
- If we decide to change the type, we'd have to change it in many places. While this is not the reader's problem, it's still annoying on the programmer's end.
- What about consistency? Nothing else in ``rmlint`` uses this, but uses the central approach. Putting it in headers also has the advantage of being able to write more about the type with traditional comments.

Also not related to the ``typedef``-question: It may be worthwhile of using another name than the rather generic ``Path``. All types in ``rmlint`` are prefixed with ``Rm`` so the reader knows that it's an internal type. Maybe ``Path`` should be renamed to ``FtsPath`` in this case.

mfwitten:
I'm probably going to rebase this series atop another, and I might end up doing some more centralization of `Path` (or&nbsp;whatever) as part of doing so.

Until then, here are my replies (which are meant to be intellectually exploratory rather than argumentative):

Why not just write the type directly?
-------------

The issues arises when using a derived type, such as an array of arrays of `char` (that is, an array of pointers to `char`; that is, a pointer to a pointer of `char`). Consider the change made in commit 167562db61300101379f5372a8a2034447ebcfab:

    -    char **path_vec = g_malloc0(sizeof(char *) * (path_count + 1));
    +    typedef const char *Path;
    +    Path *const path_vec = g_malloc0(sizeof(Path) * (path_count + 1));

The original line is pretty idiomatic; unless you *already* know what's going on, it's not quite clear at first glance what `path_vec` is really supposed to represent, namely a (null-terminated) array of path strings. The new code makes this purpose much more explicit, even with the added complexity of `const` discipline; and, once `Path` has been used to elucidate this complex case, it makes sense to continue using `Path` for the simpler cases (e.g., `const Path p = path_vec[4];` instead of `const char *const p = path_vec[4];`).

The added clarity is perhaps more apparent here:

    -    FTS *ftsp = fts_open((char *[2]){rmpath->path, NULL}, fts_flags, NULL);
    +    typedef const char *Path;
    +    FTS *ftsp = fts_open((const Path[2]){rmpath->path, NULL}, fts_flags, NULL);

In order to achieve the same result without `typedef`, one would need to write (and read!):

    FTS *ftsp = fts_open((const char *const[2]){rmpath->path, NULL}, fts_flags, NULL);

That's not terribly hard to deal with, but it's definitely harder (of course, all the `const` qualification here is not necessary, but I like to think it helps better document what's going on. After all, once again, the program is meant to be read not only by a compiler, but also by fellow humans, which brings us to the next point.

What's wrong with just looking it up once in a centralized location?
------------------

Well, nothing is wrong with it, per se, especially when one is steeped in a particular codebase (or subset therein). However, when one doesn't have a good working knowledge of the codebase in question, it can indeed be quite laborious to resolve a question whose answer requires looking through non-local code; that's why it's nice to be able to declare a quick alias right-then-and-there.

And, that's exactly what it is: A quick alias. It's not a type unto itself, which brings us to the next point.

Isn't it better to have a centralized location for the abstraction?
-----------------

The problem is that `Path` is a [*leaky* abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction); even if you *did* declare it in one place, you'd probably still have to go through a lot of sprawling code that acceses each such object as a good old-fashioned pointer to `char`, including pointer arithmetic.

If you wanted to avoid that, then you'd have to define a whole bunch of fake accessor/conversion abstractions, and for what? It's *always* going to be a C string; it's just that sometimes you want to relay such information as "We're dealing with constant C strings here" and "These strings are meant to be interpreted as paths."

Unless your language offers mechanisms to ease such abstractions (as in C++), it's just not worth the trouble, and if you're not going to stop the leaks (in the abstraction), then it's pretty pointless.

What about naming?
-------------

The local-declaration avoids the need to convey that `Path` is an internal type, and its local nature reduces the need to find some kind of consistency with the rest of the codebase. This is especially the case when it is local to a block, but also even within the file scope of a translation unit (a&nbsp;`*.c`&nbsp;file); for instance, consider the following function from `lib/cmdline.c`:

    static void rm_cmd_show_version(void)  {
        ...
    }

Why is that function namespaced? It should probably just be called `show_version()`; a non-namespaced function just means that it's internal to the translation unit.* The same logic applies to tools like `Path` (though, I do admit to placing it in a header file as well, mainly because it can be redeclared without trouble).

Anyway, I suppose our discussion here disproves me; my usage of `typedef` has been demonstrably more irritating to another reader, rather than useful.

&nbsp;

---------------------

\* While there do exist non-internal functions that are also non-namespaced (e.g., `printf`), there's just not much of a chance that they will cause any serious confusion: A fellow reader might search the translation unit for a definition, and thereby quickly realize that it must be external&mdash;and such a fact really should be documented by the `#include` directives; I've always liked the idea that every `#include` should be associated with [a comment that provides] a list of identifiers that are being "imported".

mfwitten:
In the latest push, I've simply stripped out the `typedef` tomfoolery.

sahib:
Thanks for changing that. Looks good to me & merged. :smile:

PR #317, opened 2018-10-04
Forbid `--replay' with `--dedupe' or `--is-reflink'

mfwitten:
This series documents and improves the usage of `RmCfg::path_count`, and then fixes a related programming mistake that manifests when mixing the aforementioned options. As an aside, the `gettext` files are also updated.


sahib:
Thanks. Looks good to me.

PR #319, opened 2018-10-16
Clean up path processing

mfwitten:
This series entails cleanup and refactoring, which clarify the code and improve the handling of errors in user input and runtime. Here are the main commits:

    4 [HEAD] lib/cmdline.c: Refactor `rm_cmd_set_paths*()'
    3 lib/cfg.c -> lib/cfg-funcs.h: Refactor functions
    2 Miscellaneous improvements
    1 [develop]

Each of those is composed of finer-grained commits:

    4 [HEAD] lib/cmdline.c: Refactor `rm_cmd_set_paths*()'
    |\  
    | o lib/cmdline.c: INLINE `rm_cmd_set_paths()'
    | o lib/cmdline.c: Tell compiler to optimize for the value of `replay'
    | o lib/cmdline.c: Replace `strcmp()' with direct character comparisons
    | o lib/cmdline.c: Abstract out `rm_cmd_set_paths_from_cmdline()'
    | o lib/cfg.h: Use `RmCfg::path_count' more purposefully
    | o lib/cmdline.c: Move `cfg->replay' and `cfg->path_count++' out of loops
    | o lib/cfg-funcs.h: Move `cfg->replay' and `cfg->path_count++'
    | o lib/cmdline.c: Free each command-line path as it is processed.
    | o lib/cmdline.c: Use a pointer rather than an index integer for looping
    | o lib/cmdline.c: Indent code to reflect the fact that it's in a block
    | o lib/cmdline.c: Move constant loop condition (`paths') out of the loop
    | o lib/cmdline.c: Consolidate the logic around whether to read paths from `stdin'
    | o lib/cfg.h: Remove member `read_stdin'
    | o lib/cmdline.c: Add error handling to `rm_cmd_set_paths_from_stdin()'
    | o lib/cmdline.c: Abstract out struct `rm_cmd_set_paths_vars'
    | o lib/cmdline.c: Replace `RmSession' with `RmCfg' in `rm_cmd_set_paths*()'
    | o lib/cmdline.c: Simplify size in call to `malloc()'
    | o lib/cmdline.c: Move `rm_cmd_read_paths_from_stdin()'
    |/  
    3 lib/cfg.c -> lib/cfg-funcs.h: Refactor functions
    |\  
    | o lib/cfg-funcs.h: Insert `g_assert(cfg);' into `rm_cfg_set_default()'
    | o lib/cfg.c -> lib/cfg-funcs.h: Make functions inlinable
    | |\  
    | | o lib/cfg-funcs.h: Complete `git mv lib/cfg.c lib/cfg-funcs.h'
    | | o lib/cfg.c: git mv lib/cfg.c lib/cfg-funcs.h
    | |/  
    | o lib/cfg.*: rm_cfg_prepend_path(): swap parameters `preferred' and `path'
    | o lib/cfg.*: Rename `rm_cfg_add_path' to `rm_cfg_prepend_path()'
    | o lib/path.h: s/treat_as_single_vol/single_volume/ in `RmPath'
    | o lib/path.h: s/idx/index/ in `RmPath'
    | o lib/cfg.*: Refactor `rm_cfg_add_path()'
    | o lib/cmdline.c: Test parameters of `rm_cmd_parse_replay()' with `g_assert()'
    | o lib/cmdline.c: return the status of `rm_cfg_add_path()'
    | o lib/cfg.*: `rm_cfg_add_path()' now returns `bool'
    | o lib/cfg.*: Clean up `rm_cfg_free_paths()'
    |/  
    2 Miscellaneous improvements
    |\  
    | o lib/cmdline.c: Fix error message
    | o Use explicit types for some integer constants
    |/  
    1 [develop]


lib/cmdline.c
```diff
@@ -1167,42 +1145,130 @@ static bool rm_cmd_set_cmdline(RmCfg *cfg, int argc, char **argv) {
...
+static INLINE
+void rm_cmd_set_paths_from_cmdline(
+    rm_cmd_set_paths_vars *const v,
+    const bool replay
+) {
+    g_assert(v);
+    g_assert(v->paths);
+
+    for(char *path, **list = v->paths; (path = *list); ++list) {
+        if(path[0] == '-' && path[1] == 0) {
```
sahib: Please revert that to ``strcmp`` if you don't have numbers that prove otherwise. That is just over-optimization for me and does not help readability.

lib/cmdline.c
```diff
@@ -1167,42 +1145,130 @@ static bool rm_cmd_set_cmdline(RmCfg *cfg, int argc, char **argv) {
...
+    g_assert(v);
+    g_assert(v->cfg);
+
+    RmCfg *const cfg = v->cfg;
+    const bool is_prefd = v->stdin_paths_preferred;
+
+    char delim = v->null_separated ? 0 : '\n';
+
+    size_t buf_len;
+    char *path_buf = malloc(buf_len = PATH_MAX);
```
sahib: Please revert that. It's easy enough to miss that assignment. I generally prefer direct initialization together with declaration.

sahib (changes requested):

lib/cmdline.c
```diff
@@ -1165,11 +1165,11 @@ void rm_cmd_set_paths_from_cmdline(
     g_assert(v->paths);
 
     for(char *path, **list = v->paths; (path = *list); ++list) {
-        if(strcmp(path, "-") == 0) {
+        if(path[0] == '-' && path[1] == 0) {
```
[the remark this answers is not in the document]
mfwitten: The following reply is taken from the log message of the updated commit.

-----------

The way `rmlint` currently works is that it produces a shell script that sometimes performs its work by running `rmlint` many times with varying arguments; `rmlint` can be run a large number of times, and each invocation involves argument processing whereby each path argument undergoes multiple char-by-char comparisons in search of a match for a special, well-known value (e.g., "-" or "//").

Therefore, this commit optimizes the process by using tailored, idiomatic C-string calculations in lieu of invoking a generic, open-ended, arbitrary comparison function; this commit replaces the Standard C Library's `strcmp()` with boolean operators and character comparisons.

The calculations are idiomatic in that they are a familiar way of processing strings, but are somewhat particular to C code. To aid reading this code, comments have been embedded to clarify the purpose, as in the following:

    if(/* path is "-" */ path[0] == '-' && path[1] == 0)

It's true that an optimizing compiler might perform this substitution automatically, or it might even make a better, platform-specific optimization. However, this seems unlikely based on the following rudimentary benchmarks; first define some Bash functions:

    # Copy this script with the following command line, where "$THIS"
    # expands to the revision of this commit log message:
    #
    #   git log -1 --format=%b "$THIS" |
    #     sed -n '/^  #-/,/^  #-/p' | cut -c3- >/tmp/script.sh
    #
    #----------------------------------------------------------------
    make_test()
    {
    	# Note that the body of this function is
    	# explicitly indented with tab characters
    	# for the benefit of the "<<-" operator.

    	local name="$1" expression="$2"

    	cat >"$name".c <<-EOF
    	#include <stdbool.h> // bool
    	#include <string.h>  // strcmp

    	#define ITERATIONS 10000000 // 10 million

    	int main()
    	{
    	    int result = 0;
    	    char string[3] = {0};
    	    const char *const volatile pointer = string;
    	    bool toggle = 0;
    	    for (long i = 0; i < ITERATIONS; ++i) {
    	        string[0] = string[1] = '/' + toggle;
    	        toggle = !toggle;
    	        const char *const s = pointer;
    	        result += $expression;
    	    }
    	    return result;
    	}
    	EOF
    }

    build()
    {
      local name="$1"; shift
      gcc -std=c99 -pedantic -Wall -static -fno-plt -fno-pie "$@" \
        "$name".c -o "$name"
    }

    build_all() { build strcmp "$@" && build equals "$@"; }

    run() { time -p ./"$1"; }

    benchmark()
    {
      echo =========== "$@"
      run strcmp
      echo -----------
      run equals
    }

    build_and_benchmark()
    {
      build_all "$@" || return 1
      benchmark "$@"; return 0
    }
    #----------------------------------------------------------------

Then, create and run some benchmarks:

    make_test strcmp 'strcmp(s,"//")'
    make_test equals "(s[0] == '/' && s[1] == '/' && s[2] == 0)"

    gcc --version | head -n 1
    build_and_benchmark -O0
    build_and_benchmark -Og
    build_and_benchmark -O1
    build_and_benchmark -O2

All together, this produces the following output on my [very old and slow] system; at each optimization level, timing information is provided first for `strcmp()` and then for the manual calculation.  Perversely, optimization seems to make `strcmp()` slower, not faster; in contrast, optimization drastically improves the performance of manual calculation, which is also always faster than using `strcmp()`.

    gcc (GCC) 7.3.1 20180312
    =========== -O0
    real 0.35
    user 0.35
    sys 0.00
    -----------
    real 0.24
    user 0.24
    sys 0.00
    =========== -Og
    real 0.58
    user 0.57
    sys 0.00
    -----------
    real 0.15
    user 0.15
    sys 0.00
    =========== -O1
    real 0.60
    user 0.59
    sys 0.00
    -----------
    real 0.16
    user 0.16
    sys 0.00
    =========== -O2
    real 0.58
    user 0.57
    sys 0.00
    -----------
    real 0.07
    user 0.07
    sys 0.00

Assembler code for the unoptimized case can be produced with the following:

    build_all -S -O0
    cat strcmp # review
    cat equals

Here is the meat of the comparison:

             strcmp                    equals
    ------------------------- | -------------------------
       .LC0:                  |  movl    -20(%ebp), %eax
       .string "//"           |  movzbl  (%eax), %eax
       ...                    |  cmpb    $47, %al           // s[0] == '/'
       pushl   $.LC0          |  jne     .L3
       pushl   -20(%ebp)      |  movl    -20(%ebp), %eax
       call    *strcmp@GOT    |  addl    $1, %eax
                              |  movzbl  (%eax), %eax
                              |  cmpb    $47, %al           // s[1] == '/'
                              |  jne     .L3
                              |  movl    -20(%ebp), %eax
                              |  addl    $2, %eax
                              |  movzbl  (%eax), %eax
                              |  testb   %al, %al           // s[2] == 0
                              |  jne     .L3
                              |  movl    $1, %eax
                              |  jmp     .L4
                              |  .L3:
                              |  movl    $0, %eax

As you can see, one is calling a function, while the other is performing the manual calculations. Now, consider the optimized version:

    build_all -S -O2
    cat strcmp
    cat equals

This yields:

             strcmp                    equals
    ------------------------- | -------------------------
     .LC0:                    |  movl    -20(%ebp), %ebx
     .string "//"             |  cmpb    $47, (%ebx)
     ...                      |  jne     .L2
     movl    -36(%ebp), %esi  |  cmpb    $47, 1(%ebx)
     movl    $.LC0, %edi      |  jne     .L2
     ...                      |  cmpb    $1, 2(%ebx)
     movl    $3, %ecx         |
     repz cmpsb               |
     seta    %cl              |
     setb    %bl              |
     subl    %ebx, %ecx       |
     movsbl  %cl, %ecx        |

There is no longer a call to `strcmp()`, but the compiler has substituted what is essentially an exact replica of `strcmp()` in terms of x86 assember code; the `repz cmpsb` iterates through 2 strings in memory, comparing the bytes, stopping when there is a mismatch. Strangely, one would think that a native x86 instruction sequence might be inherently faster, but the benchmarks reveal that the naive integer comparisons are much faster. I suspect the reason for this is 2-fold:

* Perhaps, `repz cmpsb` is performing a lot more memory access; in contrast, each naive integer comparison instruction can encode a comparand directly (e.g., the number 47, which is ASCII for '/').

* Maybe, `repz cmpsb` cannot benefit from instruction pipelining as readily as separate integer comparisons.

Perhaps the GNU GCC team should re-consider how `strcmp()` is optimized.

Anyway, the results are clear: It's better to do these small calculations by hand, rather than rely on the compiler to do something magical.

lib/cmdline.c
mfwitten: In addition to [the other comment](https://github.com/sahib/rmlint/pull/319#discussion_r250416132), I do not think there is such a thing as "over-optimization"; the problem with optimization is not that it's over done, but rather as you imply: It might damage some other aspect of the program, such as readability.

Then again, why should readability be the winning aspect? Perhaps `strcmp()` is *over-readablized*, to the unnecessary detriment of optimization. Indeed, a C-string is explicitly an array of characters, and inspecting individual characters (especially when looking for a very short string) is quite normal&mdash;it's one of the killer features C was designed to have.

If there is a need to improve readability, then other means should be employed before sacrificing run-time efficiency: Abstraction facilities (such as macros or inlinable functions or structs), or even just plain, free-form, natural human descriptions in the form of embedded comments.

I mean is something like `strcmp(x,"y") == 0` actually *that* readable? Sure, a competent C programmer should be familiar with it, but he should also be familiar with indexing individual characters, too; both are equally readable from that perspective, so one might as well choose the solution that has better performance.

mfwitten:
The original patch series has essentially been updated as per the following diff:

    diff --git a/lib/cfg-funcs.h b/lib/cfg-funcs.h
    index 9b218c31..c2e4f000 100644
    --- a/lib/cfg-funcs.h
    +++ b/lib/cfg-funcs.h
    @@ -121,8 +121,10 @@ bool rm_cfg_prepend_json(
                 real_path,
                 cfg->path_count++,
                 false /* not preferred */
    -        ); return true;
    -    } return false;
    +        );
    +        return true;
    +    }
    +    return false;
     }
     
     #define PREPEND_TO(list)    \
    @@ -149,8 +151,10 @@ bool rm_cfg_prepend_path(
             } else {
                 PREPEND_TO(cfg->paths);
                 ++(cfg->path_count);
    -        } return true;
    -    } return false;
    +        }
    +        return true;
    +    }
    +    return false;
     }
     
     #undef PREPEND_TO
    @@ -159,9 +163,9 @@ static INLINE
     void rm_cfg_free_paths(RmCfg *const cfg) {
         g_assert(cfg);
         g_slist_free_full(cfg->paths, (GDestroyNotify)rm_path_free);
    -    cfg->paths = 0;
    +    cfg->paths = NULL;
         g_slist_free_full(cfg->json_paths, (GDestroyNotify)rm_path_free);
    -    cfg->json_paths = 0;
    +    cfg->json_paths = NULL;
     }
     
     #endif /* end of include guard */
    diff --git a/lib/cmdline.c b/lib/cmdline.c
    index 56d07a8e..a3ab75f9 100644
    --- a/lib/cmdline.c
    +++ b/lib/cmdline.c
    @@ -1074,12 +1074,25 @@ static gboolean rm_cmd_parse_rankby(_UNUSED const char *option_name,
     
     static gboolean rm_cmd_parse_replay(_UNUSED const char *option_name,
                                         const gchar *path, RmSession *session,
    -                                    _UNUSED GError **error) {
    +                                    GError **error) {
         g_assert(session);
         g_assert(session->cfg);
         session->cfg->replay = true;
         session->cfg->cache_file_structs = true;
    -    return rm_cfg_prepend_json(session->cfg, path);
    +
    +    if(rm_cfg_prepend_json(session->cfg, path)) {
    +        return true;
    +    }
    +
    +    g_assert(error);
    +    g_assert(*error == NULL);
    +
    +    g_set_error(
    +        error, RM_ERROR_QUARK, 0,
    +        _("Failed to include this replay file: %s"), path
    +    );
    +
    +    return false;
     }
     
     static gboolean rm_cmd_parse_equal(_UNUSED const char *option_name,
    @@ -1145,7 +1158,7 @@ static bool rm_cmd_set_cmdline(RmCfg *cfg, int argc, char **argv) {
         return true;
     }
     
    -typedef struct rm_cmd_set_paths_vars {
    +typedef struct RmCmdSetPathVars {
         RmCfg *const cfg;
         char **const paths;
         unsigned int index;
    @@ -1154,23 +1167,21 @@ typedef struct rm_cmd_set_paths_vars {
         bool stdin_paths_preferred;
         const bool null_separated;
         bool all_paths_valid;
    -} rm_cmd_set_paths_vars;
    +} RmCmdSetPathVars;
     
     static INLINE
     void rm_cmd_set_paths_from_cmdline(
    -    rm_cmd_set_paths_vars *const v,
    +    RmCmdSetPathVars *const v,
         const bool replay
     ) {
         g_assert(v);
         g_assert(v->paths);
     
         for(char *path, **list = v->paths; (path = *list); ++list) {
    -        if(path[0] == '-' && path[1] == 0) {
    +        if(/* path is "-" */ path[0] == '-' && path[1] == 0) {
                 v->read_stdin = true;
    -            /* remember whether to treat stdin paths as preferred paths */
                 v->stdin_paths_preferred = v->is_prefd;
    -        } else if(path[0] == '/' && path[1] == '/' && path[2] == 0) {
    -            /* the '//' separator separates non-preferred paths from preferred */
    +        } else if(/* path is "//" */ path[0] == '/' && path[1] == '/' && path[2] == 0) {
                 v->is_prefd = !v->is_prefd;
             } else {
                 v->all_paths_valid &= rm_cfg_prepend_path(
    @@ -1184,7 +1195,7 @@ void rm_cmd_set_paths_from_cmdline(
     
     static INLINE
     bool rm_cmd_set_paths_from_stdin(
    -    rm_cmd_set_paths_vars *const v,
    +    RmCmdSetPathVars *const v,
         const bool replay
     ) {
         g_assert(v);
    @@ -1195,8 +1206,8 @@ bool rm_cmd_set_paths_from_stdin(
     
         char delim = v->null_separated ? 0 : '\n';
     
    -    size_t buf_len;
    -    char *path_buf = malloc(buf_len = PATH_MAX);
    +    size_t buf_len = PATH_MAX;
    +    char *path_buf = malloc(buf_len);
         if(!path_buf) {
             rm_log_perror(_("Failed to allocate memory"));
             return false;
    @@ -1243,7 +1254,7 @@ bool rm_cmd_set_paths(RmCfg *const cfg, char **const paths) {
     
         const bool replay = cfg->replay;
         const bool read_stdin0 = cfg->read_stdin0;
    -    rm_cmd_set_paths_vars v = {
    +    RmCmdSetPathVars v = {
             .cfg = cfg,
             .paths = paths,
             .index = cfg->path_count,
    diff --git a/lib/path-funcs.h b/lib/path-funcs.h
    index 5599387b..1f46ac0b 100644
    --- a/lib/path-funcs.h
    +++ b/lib/path-funcs.h
    @@ -39,7 +39,7 @@
     #endif
     
     #include "path.h"       // RmPath
    -#include "config.h"     // _, rm_log_warning_line, rm_log_perrorf
    +#include "config.h"     // _, rm_log_warning_line
     
     static INLINE
     void rm_path_free(RmPath *const p) {
    @@ -63,7 +63,9 @@ bool rm_path_is_real(
         rm_log_warning_line(
             _("Can't get real path for directory or file \"%s\": %s"),
             path, strerror(errno)
    -    ); return false;
    +    );
    +
    +    return false;
     }
     
     #if HAVE_FACCESSAT
    @@ -79,8 +81,10 @@ bool rm_path_is_accessible(const char *const path) {
             rm_log_warning_line(
                 _("Can't open directory or file \"%s\": %s"),
                 path, strerror(errno)
    -        ); return false;
    -    } return true;
    +        );
    +        return false;
    +    }
    +    return true;
     }
     
     #undef NOT_ACCESSIBLE
    @@ -109,8 +113,10 @@ bool rm_path_is_file(const char *const path) {
             rm_log_warning_line(
                 _("Could not get metadata for path \"%s\": %s"),
                 path, strerror(errno)
    -        ); return false;
    -    } return S_ISREG(s.st_mode);
    +        );
    +        return false;
    +    }
    +    return S_ISREG(s.st_mode);
     }
     
     static INLINE
    diff --git a/po/de.po b/po/de.po
    index feb59eb5..e4d23497 100644
    --- a/po/de.po
    +++ b/po/de.po
    @@ -7,7 +7,7 @@ msgid ""
     msgstr ""
     "Project-Id-Version: rmlint 2.0.0\n"
     "Report-Msgid-Bugs-To: \n"
    -"POT-Creation-Date: 2018-10-15 04:46+0000\n"
    +"POT-Creation-Date: 2019-01-21 21:40+0000\n"
     "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
     "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
     "Language-Team: LANGUAGE <LL@li.org>\n"
    @@ -1028,6 +1028,11 @@ msgstr ""
     msgid "No stamp file at `%s`, will create one after this run."
     msgstr ""
     
    +#: lib/cmdline.c
    +#, c-format
    +msgid "Failed to include this replay file: %s"
    +msgstr ""
    +
     #: lib/path-funcs.h
     #, c-format
     msgid "Invalid path \"%s\""
    diff --git a/po/es.po b/po/es.po
    index d0a4c46e..fd81c42d 100644
    --- a/po/es.po
    +++ b/po/es.po
    @@ -8,7 +8,7 @@ msgid ""
     msgstr ""
     "Project-Id-Version: rmlint 2.4.0\n"
     "Report-Msgid-Bugs-To: \n"
    -"POT-Creation-Date: 2018-10-15 04:46+0000\n"
    +"POT-Creation-Date: 2019-01-21 21:40+0000\n"
     "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
     "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
     "Language-Team: LANGUAGE <LL@li.org>\n"
    @@ -1005,6 +1005,11 @@ msgstr ""
     msgid "No stamp file at `%s`, will create one after this run."
     msgstr ""
     
    +#: lib/cmdline.c
    +#, c-format
    +msgid "Failed to include this replay file: %s"
    +msgstr ""
    +
     #: lib/path-funcs.h
     #, c-format
     msgid "Invalid path \"%s\""
    diff --git a/po/fr.po b/po/fr.po
    index 2b2c58c1..ea5b2db4 100644
    --- a/po/fr.po
    +++ b/po/fr.po
    @@ -7,7 +7,7 @@ msgid ""
     msgstr ""
     "Project-Id-Version: rmlint 2.0.0\n"
     "Report-Msgid-Bugs-To: \n"
    -"POT-Creation-Date: 2018-10-15 04:46+0000\n"
    +"POT-Creation-Date: 2019-01-21 21:40+0000\n"
     "PO-Revision-Date: 2014-12-02 13:37+0100\n"
     "Last-Translator: F. <contact@f.0x2501.org>\n"
     "Language-Team: none\n"
    @@ -1001,6 +1001,11 @@ msgstr ""
     msgid "No stamp file at `%s`, will create one after this run."
     msgstr ""
     
    +#: lib/cmdline.c
    +#, c-format
    +msgid "Failed to include this replay file: %s"
    +msgstr ""
    +
     #: lib/path-funcs.h
     #, c-format
     msgid "Invalid path \"%s\""
    diff --git a/po/rmlint.pot b/po/rmlint.pot
    index 94b92658..348f86fa 100644
    --- a/po/rmlint.pot
    +++ b/po/rmlint.pot
    @@ -8,7 +8,7 @@ msgid ""
     msgstr ""
     "Project-Id-Version: rmlint 2.7.0\n"
     "Report-Msgid-Bugs-To: \n"
    -"POT-Creation-Date: 2018-10-15 04:46+0000\n"
    +"POT-Creation-Date: 2019-01-21 21:40+0000\n"
     "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
     "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
     "Language-Team: LANGUAGE <LL@li.org>\n"
    @@ -950,6 +950,11 @@ msgstr ""
     msgid "No stamp file at `%s`, will create one after this run."
     msgstr ""
     
    +#: lib/cmdline.c
    +#, c-format
    +msgid "Failed to include this replay file: %s"
    +msgstr ""
    +
     #: lib/path-funcs.h
     #, c-format
     msgid "Invalid path \"%s\""

RayOei:
@vassilit Are you planning to do something with this one (I noticed your comment) or should we close it? It seems it never got resolved when Sahib was still in control and the code has changed and this branch has conflicts and there was also a clear disagreement on the necessity of this change... soo... your thoughts?

vassilit:
> @vassilit Are you planning to do something with this one (I noticed your comment) or should we close it? It seems it never got resolved when Sahib was still in control and the code has changed and this branch has conflicts and there was also a clear disagreement on the necessity of this change... soo... your thoughts?

There are some interesting ideas, I would like to keep access to those commits to deal with them in the future.
I'm afraid to close this PR and the branch get deleted automatically.

vassilit:
I was thinking about cherry-picking the changes that we find reasonable in this PR and merge it. That way we can attribute the work to its original author.
If you think otherwise, feel free to close this PR.

RayOei:
Aren't the authors saved, regardless? But I am fine either way. As long as there is _some_ resolution in the near future. But that is maybe my OCD-ish nature to clean up things which are sitting somewhere for ages 🤓 
This repo is not that bad anyway, compared to some others 😀

PR #331, opened 2019-01-10
Fix nosetests search.

e7appew:
Avoid confusing error message about nosetests not being found when
there is a test failure.


sahib:
Thanks @e7appew.

PR #332, opened 2019-01-11
Fix support for pre-SSE4.2 CPUs.

e7appew:
When compiled with -msse4.2 optimisation applied across the board,
'rmlint -g' terminates with the SIGILL signal on pre-SSE4.2 platforms.

It seems like GCC 8.2.0 optimises the floor function/macro call from
lib/formats/progressbar.c with illegal instructions. With Clang, even
more such optimisations occur.

Use function attributes to enable SSE4.2 feature only where required,
so that both types of CPU are fully supported.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -663,8 +663,6 @@ else:
 
 # check _mm_crc32_u64 (SSE4.2) support:
 conf.check_mm_crc32_u64()
-if conf.env['HAVE_MM_CRC32_U64']:
-    conf.env.Append(CCFLAGS=['-msse4.2'])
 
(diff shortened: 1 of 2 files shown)


sahib:
Thank you @e7appew.

> When compiled with -msse4.2 optimisation applied across the board,
'rmlint -g' terminates with the SIGILL signal on pre-SSE4.2 platforms.

Ah, crap. I remember having that issue already on Arch, where the packager had a SSE4.2 capable machine, but most users did not. Glad that you fixed it.

> Use function attributes to enable SSE4.2 feature only where required,
so that both types of CPU are fully supported.

I was not aware of that feature. Thanks again.

PR #350, opened 2019-02-07
Fix and add a few FR translations

bzg:


sahib:
Thanks a lot!

bzg:
Thanks to you! rmlint is great.

PR #358, opened 2019-04-20
Teach rmlint to detect reflinks with > 2GB files

reedriley:
The `rm_offset_get_fiemap` function took a 32-bit signed integer describing the file offset; but the [kernel data structure is defined as a 64-bit unsigned integer](https://www.kernel.org/doc/Documentation/filesystems/fiemap.txt).  This caused rmlint to fail to read the fiemap for files larger than 2 GB.

In the process of debugging and fixing the issue; I ran into some compilation logs in the `_RM_OFFSET_DEBUG` logging, so I fixed those while I was at it.


sahib:
Thank you. Just tested it and it seems to work. :+1:

PR #387, opened 2019-11-30
added another check for HAVE_XATTR in xattr.c

pianojockl:
without this check for HAVE_XATTR, build fails for systems that don't have sys/xattr.h
I'm working on a nice FreeBSD port and stumbled upon this.


sahib:
Thanks @pianojockl.

PR #391, opened 2019-12-31
Remove use of GLib API calls deprecated since version 2.62.

e7appew:
Replace deprecated GTimeVal with GDateTime equivalent API calls.

Only if you no longer need to support versions of GLib older than this.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -611,7 +611,7 @@ conf.check_pkgconfig('0.15.0')
 
 # Pkg-config to internal name
 conf.env['HAVE_GLIB'] = 0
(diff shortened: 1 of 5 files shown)


sahib:
This already has been fixed in the ``develop`` branch (in 49385019ce02203027ff2d414fb650df40fbf374, d3f66aee9934b798d16b9b0de74f9b5f65553e5e and 94049c3e2946fdcb1fa08172b285c24026ec0fff)

sahib:
Also, the fixes there still support older versions. At least TravisCI requires us to do so (for now), but there are also still plenty people that use ancient software.

sahib:
Still thank you for the PR!

PR #392, opened 2020-01-01
Shellcheck cleanup

rpdelaney:
This fixes [shellcheck](https://shellcheck.net) linter errors and fixes a problem with how `iconv` is being invoked, which was referenced in the comments.

shellcheck's [SC2094](https://github.com/koalaman/shellcheck/wiki/SC2094) explains the problem with the code as it was.

### DIFF
--- a/.format.sh
+++ b/.format.sh
@@ -3,11 +3,11 @@
 FILES=$(find src lib -iname '*.[ch]')
 
 # Format the files according to the style defined in .clang-format
-clang-format -i -style=file "${FILES}" || (echo 'clang-format failed'; exit 1);
+clang-format -i -style=file ${FILES} || (echo 'clang-format failed'; exit 1);
 
 # clang-format sometimes adds windows like line-endins (\r\n).
 # Fix this by issueing dos2unix on every file.
-dos2unix --quiet "${FILES}" || (echo 'dos2unix failed'; exit 2);
+dos2unix --quiet ${FILES} || (echo 'dos2unix failed'; exit 2);
 
 # Some bad editors might save files with a bad encoding.
 # Since a few source code files contain unicode, we fix this with iconv.
--- a/.format.sh
+++ b/.format.sh
@@ -3,16 +3,16 @@
 FILES=$(find src lib -iname '*.[ch]')
 
 # Format the files according to the style defined in .clang-format
-clang-format -i -style=file ${FILES} || (echo 'clang-format failed'; exit 1);
+clang-format -i -style=file "${FILES}" || (echo 'clang-format failed'; exit 1);
 
 # clang-format sometimes adds windows like line-endins (\r\n).
-# Fix this by issueing dos2unix on every file. 
-dos2unix --quiet ${FILES} || (echo 'dos2unix failed'; exit 2);
+# Fix this by issueing dos2unix on every file.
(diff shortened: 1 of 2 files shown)


rpdelaney:
Looking at this again, I think the change from `${FILES}` to `"${FILES}"` will likely break this in a different way. To avoid problems with whitespace, globs and so on, an entirely different approach is necessary. I will fix that up tomorrow.

Happy new year :)

lib/formats/sh.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 PROGRESS_CURR=0
-PROGRESS_TOTAL=0                           
+PROGRESS_TOTAL=0
```
sahib: Those spaces here are (embarrassingly) needed, since the remainder after the ``=`` is replaced with the actual number. Having those spaces in allows us to do this in-place without shifting the rest of the file.

.format.sh
```diff
@@ -3,16 +3,21 @@
 FILES=$(find src lib -iname '*.[ch]')
 
 # Format the files according to the style defined in .clang-format
-clang-format -i -style=file ${FILES} || (echo 'clang-format failed'; exit 1);
+clang-format -i -style=file "${FILES}" || (echo 'clang-format failed'; exit 1);
```
sahib: That probably breaks in case of several files. Maybe using ``xargs`` is a better option here.

lib/formats/sh.sh
```diff
@@ -255,19 +255,19 @@ remove_cmd() {
     echo "${COL_YELLOW}Deleting: ${COL_RESET}$1"
     if original_check "$1" "$2"; then
         if [ -z "$DO_DRY_RUN" ]; then
-            if [ ! -z "$DO_KEEP_DIR_TIMESTAMPS" ]; then
-                touch -r "$(dirname $1)" "$STAMPFILE"
+            if [ -n "$DO_KEEP_DIR_TIMESTAMPS" ]; then
+                touch -r "$(dirname "$1")" "$STAMPFILE"
```
sahib: Do nested quotations like this work? I have some doubts here. :)

sahib:
Hi @rpdelaney,

thanks for your contribution. I left a few comments with concerns.

Happy new year to you too!

.format.sh
rpdelaney: Yes, I mentioned this in my comment

lib/formats/sh.sh
rpdelaney: What is the source of your doubt? The quotes are required inside a subshell because the substitution of `$1` is vulnerable to error from filenames with whitespace, newlines etc.

sahib: Hmm, I thought this would break due the surrounding quotes, but I had a quick try in the shell and it worked. So consider those doubts as cleared.

sahib:
Merged this PR - the actual issue of the directory quoting is more urgent to fix than the small issue we still had open in ``.format.sh``. Thanks @rpdelaney.

rpdelaney:
Hey, sorry for letting this get so stale. I'll leave the issue bookmarked and hopefully come back to it. It's a neat tool and I'm still using it, so it's in the back of my mind if not on top of it :-) Thank you

PR #411, opened 2020-05-23
fix unnecessary STAMPFILE creation in rmlint.sh

JLuszawski:
when DO_KEEP_DIR_TIMESTAMPS was not set the STAMPFILE was created but not removed


sahib:
Thanks @JLuszawski!

PR #418, opened 2020-06-11
Escape contents of 'parent_path' JSON field

EvilMcJerkface:
Just ran:

`rmlint . -D -T minimaldirs`

and some of the JSON produced was broken, namely any time a 'part_of_directory' record was generated for a path that had something goofy like one double-quote in it, the 'path' field was escaped properly but the 'parent_path' field was not.  The 'path' field output is escaped via 'rm_fmt_json_key_unsafe', whereas the 'parent_path' field just uses 'rm_fmt_json_key'.  Probably both ought to use _unsafe.


sahib:
Oh, damn. Thanks for noticing, I will release a bugfix release with this patch.

PR #420, opened 2020-06-18
Pedantic edits for sh

Artoria2e5:
* Remove all invocations of echo due to the poorly-defined handling of
  backslashes (will show up in paths because user input). Now using printf.
* Do not let C handle \e, because that's non-standard.
* Remove handling of percents in SConscript and make it not hand escapes
  to C.
* Add -- to invocations of external commands to guard against potential
  option-looking filenames.
* The same stuff in sh.c.in.


lib/SConscript
```diff
@@ -50,24 +50,20 @@ def build_config_template(target, source, env):
         handle.write(config_h)
 
 
-def prepare_c_source(text, escape_percentages=True):
+def prepare_c_source(text):
     # Prepare the Python source to be compatible with C-strings
     text = text.replace('"', '\\"')
-    text = text.replace('\\n', '\\\\n')
+    text = text.replace('\\', '\\\\')
```
sahib, line 56: What's that reason for this change? It breaks compilation. The issue is also visible on Jenkins: https://travis-ci.org/github/sahib/rmlint/jobs/699617161

Artoria2e5: The idea is to don't leak any escapes to C at all. I effed up when I let the quotes run first.

sahib:
Please change the PR target to ``develop``, then I will merge it.

Artoria2e5:
Done.

PR #443, opened 2020-11-10
Don't disable fiemap when gettext is disabled

tsudoko:
fiemap does not depend on gettext.


sahib:
Oops, thanks. :smile:

Issue #448, opened 2020-12-08
Python script KeyError: 'part_of_directory'


nealmcb:
The py.py script hasn't kept up-to-date with the `-D` option and the new `part-of-directory` type.

```
$ rmlint -D -c sh:link -O py:remwithdir.py
...

$ ./remwithdir.py -n
# This is a dry run. Nothing will be modified.
[100%] Deleting empty directory: /srv/mirror/s/rmlint/.git/refs/tags
[100%] Deleting empty directory: /srv/mirror/s/rmlint/.git/objects/info
[100%] Deleting empty directory: /srv/mirror/s/rmlint/.git/branches
Traceback (most recent call last):
  File "./remwithdir.py", line 272, in <module>
    main(args, json_doc)
  File "./remwithdir.py", line 200, in main
    msg = ORIGINAL_MESSAGES[item['type']].format(c=COLORS)
KeyError: 'part_of_directory'
```

Context via pdb:

```
python -m pdb ./remwithdir.py -n
...
 File "./remwithdir.py", line 200, in main                                                                                                                                      
    msg = ORIGINAL_MESSAGES[item['type']].format(c=COLORS)                                                         
KeyError: u'part_of_directory'                                                                                                                                          
Uncaught exception. Entering post mortem debugging                                                                            
Running 'cont' or 'step' will restart the program                                                                                                                    
> /srv/mirror/s/rmlint/remwithdir.py(200)main()                                                                             
-> msg = ORIGINAL_MESSAGES[item['type']].format(c=COLORS)                                                                                                                  

(Pdb) item
{u'checksum': u'046f72840ae88b4940b342512a937f8848784a391c448d89883c025d6fcded8622e3053575a1a01b3678527bd1e5836667382799dcbbad384741191636fe3141', u'mtime': 1607292742.0291479,
u'is_original': True, u'disk_id': 46, u'parent_path': u'/srv/mirror/s/rmlint/docs', u'depth': 2, u'inode': 71931285, u'path': u'/srv/mirror/s/rmlint/docs/tutorial.rst', u'progress': 100, u'type': u'part_of_directory', u'id': 712920633, u'size': 35039}

(Pdb) item['type']
u'part_of_directory'

(Pdb) ORIGINAL_MESSAGES
{'duplicate_file': '{c[green]}Keeping original:  ', 'duplicate_dir': '{c[green]}Keeping original directory:  '}

```

sahib:
[...]

PR #449, opened 2020-12-08
handle part_of_directory in py.py

nealmcb:
Beginning of a fix to #448

This will at least allow the python script to run to completion with the `-n` option.

But since I'm not really sure how to handle `part_of_directory`, please don't use this for real!
It is untested.


sahib:
Thanks I changed it in 84b50169 so that it's safer to add new types and still not to crash in that case.

Issue #385, opened 2019-10-26
Shredder does not include manually toggled files


megitu:
After manually toggling individual files and running "Render script from Selected", these toggled files are not correctly marked in the generated script.
The generated script is always the same as in the case of "Render script from All" (and these manually toggled files are not correctly marked on the lines: "original_cmd" and "remove_cmd").

sahib:
Hello @megitu,

What version are you using? This is probably a duplicate of #375. Please check if compiling from the  ``develop`` branch fixes it for you.

megitu:
I am using 2.9.0 under Manjaro Linux.
[#375](https://github.com/sahib/rmlint/issues/375), as I understand it, is a similar case and much more completed but probably not the same.
I will try to use the develop branch fixes (if I can use it).

ecdlguy:
Hi, thank you for writing this issue... It saved me from deleting some dupes at the wrong location. 
I've noticed that toggling files in shredder is never rendered correctly in the generated script.
I'm using version version 2.8.0 "Maidenly Moose" (rev 3), though.

HaleTom:
+1 for keyboard shortcuts for both `Toggle selected` and `Toggle all`.

With a lot of dupes, it's RSI-inducing to use the moose for this :)

s-rogonov:
also it seems like I f****d up a looooot of my files =/
sad to be me

SeeSpotRun:
Hi all, sorry for very tardy buy-in to this.

Yes I can reproduce the issue as follows:
* Scan a folder using `rmlint --gui`
* Manually toggle files to tag a different file as original.
* Generate `rmlint.sh`,
* Desired result: the manually tagging of originals is reflected in the shell output script.
* Observed result: the manually tagging of originals is not reflected in the shell output script.

@sahib I can't find anything in the `--replay` logic that preserves the originals status from the json cache.  I went ahead and coded #482 to do this, but would prefer to have your feedback before merging it, in case I'm missing something.

SeeSpotRun:
This should be addressed as of https://github.com/sahib/rmlint/commit/b94e5aca2d54ca3240e5d94c342a2686bb6ab16b which has been merged into `develop` branch

Use option `--keep-cached-originals` ("For --replay runs, preserve any originals in json cache")

PR #465, opened 2021-01-14
fix issue #385

s-rogonov:
I hope I find a solution of the issue #385.

current behavior:

1. reads json properties
2. overwrites "is_original" prop to first-in-the-duplicate group

suggested behavior:

1. reads json properties
2. keeps "is_original" prop unchanged (due to gui "shredder" overwrites it to save manually toggled files)

### DIFF
--- a/lib/replay.c
+++ b/lib/replay.c
@@ -738,13 +738,14 @@ static void rm_parrot_cage_write_group(RmParrotCage *cage, GQueue *group, bool p
     for(GList *iter = group->head; iter; iter = iter->next) {
         RmFile *file = iter->data;
 
-        if(file == group->head->data
+        // just use "is_original" value from json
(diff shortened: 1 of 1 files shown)


SeeSpotRun:
Unfortunately this is causing this test to fail:  https://github.com/sahib/rmlint/blob/develop/tests/test_options/test_replay.py#L128-L149

Basically for command-line users, sometimes it's useful to re-run but with different (automated) ranking of originals.  Your proposed change breaks this.

A workaround might be to add an option `-S j` to rank originals based on their prior ranking in the `json` file.  Although it gets a bid messy if new files are added between runs (or go missing).

SeeSpotRun:
Underlying issue should be addressed by #482.

New option `--keep-cached-originals` will respect `is_original` property from json file.

PR #467, opened 2021-01-21
Clone hardlinks

### LINKED ISSUE #466: clone hardlinks: FIDEDUPERANGE returned error: (22): Invalid argument
I ran 

```term
rmlint --types="duplicates" --config=sh:handler=clone
```

and then ran the `rmlint.sh`, and it successfully combines a bunch of files, but also fails with this error on a bunch of others.  I see no obvious difference between the file names that would throw off the command line.  For example, these two files are the same:
```term
> cmp ErFLQmkWMAMTWFu.jpeg ErFLQmkWMAMTWFu.jpg
```

But rmlint can't combine them:
```term
> rmlint --dedupe ErFLQmkWMAMTWFu.jpeg ErFLQmkWMAMTWFu.jpg
ERROR: lib/session.c:331: FIDEDUPERANGE returned error: (22): Invalid argument
```

Happens in `fish` or `bash`

```term
> rmlint --version
version 2.9.0 compiled: Dec 31 2019 at [22:27:25] "Odd Olm" (rev 2)
compiled with: +mounts +nonstripped +fiemap +sha512 +bigfiles +intl +replay +xattr +btrfs-support

rmlint was written by Christopher <sahib> Pahl and Daniel <SeeSpotRun> Thomas.
The code at https://github.com/sahib/rmlint is licensed under the terms of the GPLv3.

> uname -a
Linux 5.4.0-60-generic #67-Ubuntu SMP Tue Jan 5 18:31:36 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

Ohhh, you know what it probably is?  I already merged a bunch of files as hardlinks first.  These unreflinkable files have the same inode:

```term
> ls -il ErFL*
8172271 -rw-rw-r-- 3 endolith endolith 756441 Jan 14 16:36 ErFLlquVEAEeXvX.jpg
8206761 -rw-rw-r-- 5 endolith endolith 644501 Jan 14 16:36 ErFLQmkWMAMTWFu.jpeg
8206761 -rw-rw-r-- 5 endolith endolith 644501 Jan 14 16:36 ErFLQmkWMAMTWFu.jpg
8172275 -rw-rw-r-- 3 endolith endolith  62767 Jan 14 16:36 ErFLVW3UcAETeps.jpg
```

SeeSpotRun:
Fixes #466.

This PR enables hardlink-clones to be converted to reflink-clones via a single command `rmlint --dedupe <original> <hardlink>`, while preserving file permissions and times.

Note that converting hardlink to reflink doesn't free any space, in fact it consumes a little bit for the necessary metadata.  The advantage is that it de-links the two files' metadata, so that subsequent changes to permissions, access times etc are independent for the two files.  Also modifications such as appending data to the original won't affect the reflink the way it would the hardlink.

Note that this implementation does temporarily delete the hardlink, so an untimely crash or error could see the hardlink lost (but not the original file).  The workaround would require generating a unique temporary file name for the reflink.

Happy to discuss.


SeeSpotRun:
I haven't gone any further with this because it can't be done atomically so there is some risk that the hardlink gets deleted or renamed.  I could do a bit more work to make it more robust but will wait to see if there is any more interest in #466

SeeSpotRun:
Ok managed to make it reasonably atomic by cloning to a tempfile in the target dir and then atomically renaming that over the top of the hardlink.

PR #470, opened 2021-02-11
Move all CC checks after setting CC/CXX vars

kdm9:
Hello,

While trying to package rmlint for the conda package manager, I came across the following issue: compiler paths and flags from `os.environ` are incorporated into the build environment after some checks (namely `CheckCC()`) execute. Therefore, if the default values of scons' CC etc. are non-functional (as in a cross-build environment where CC is named by host arch etc.), then these tests fail, aborting the build before paths to CC and CXX are read from environment variables. This simply shifts the block of code that overwrites scons' CC/CXX etc up before any checks are performed. The attached patch seems to allow successful compilation & install at least in Conda's build environment.

Best,
Kevin

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -605,6 +605,34 @@ conf = Configure(env, custom_tests={
     'check_sysmacro_h': check_sysmacro_h
 })
 
+#######################################################################
+#                      Compiler Checks and Flags                      #
+#######################################################################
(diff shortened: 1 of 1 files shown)


SeeSpotRun:
Thanks @kdm9

PR #474, opened 2021-03-05
Fix print of config parameters

Fat-Zer:
Otherwise ARGUMENTS.get('DEBUG') evaluated to `True` when actually `0` was passed


SeeSpotRun:
Thanks

Issue #462, opened 2021-01-04
json has no hashsum?


lockywolf:
I am trying to deduplicate remote disks, in a way that is suggested by issues https://github.com/sahib/rmlint/issues/329 and https://github.com/sahib/rmlint/issues/199

I'm running `time rmlint -g -c json:unique -mkr // /home/`.
When I am browsing the resulting json, I see no field for a hash sum. How would rmlint on a different machine, using `--replay`, find duplicates?

SeeSpotRun:
Yes, in the interest of execution time we don't complete checksums on files which diverge in the first few kb or so.  Could add an option `-c json:hash_uniques` as a work-around if you still need this.

SeeSpotRun:
Ok @lockywolf , two new options have now been added and merged into the develop branch https://github.com/sahib/rmlint/tree/develop.
With `--hash-uniques`, all found files get hashed.
With `--hash-unmatched`, only size-twins get hashed.  This is more efficient for dupe-finding, because if you only have one file that is 4,635,235,654 bytes long then it can't have any duplicates.

Also with either of these options specified, you no longer need `-c json:unique`

lockywolf:
That's going to help, thank you!

Shall the issue be closed?

SeeSpotRun:
Resolved by #479.

PR #479, opened 2021-03-16
Implement new options to fully hash unique files

SeeSpotRun:
Addresses #462

New option `--hash-uniques` means even unique files get fully hashed.  These will generally be outputted to the json report by default.  By also specifying `--xattr-write`, the checksums will be written to the files' extended attributes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format follows [keepachangelog.com]. Please stick to it.
 +### Added
 +
 +* Implement --hash-uniques option to generate full checksums of unique files too.
++* Implement --hash-unmatched option, similar to --hash-uniques but only for size twins.
 
 ## [2.10.1 Ludicrous Lemur] -- 2020-06-13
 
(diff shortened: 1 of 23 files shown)


SeeSpotRun:
@sahib I wonder if `write-unfinished` can / should be deprecated with this in place?  Unfinished checksums seem a bit flakey.  They're only really useful if another same-length file has a different unfinished checksum after exactly the same number of bytes.  I can't see where we are storing the number of bytes hashed in the json file for `--replay` so I'm not sure how it's really supposed to work.
While `--hash-uniques` is a bigger overhead for the first run, it's arguably more useful and robust than partial checksums.

sahib:
I can't find the ticket right now, but there were some issues with ``--write-unfinished`` anyways (at least in the form that most users assumed it does what ``--hash-uniques`` does). Especially since it got implicitly enabled with the xattr feature. I would vote to remove it altogether.

SeeSpotRun:
Ok have removed and also added `--hash-unmatched` which is like `--hash-uniques` but only hashes files that have one or more size twins.  This will be much more efficient in most usecases.

Issue #464, opened 2021-01-06
WARNING: Error: JSON data must be UTF-8 encoded


He1my:
rmlint is unable to read a 50M json file it has produced after scanning a large number of files 
$file large.json 
large.json: UTF-8 Unicode text

see attached file, 
[large.tar.gz](https://github.com/sahib/rmlint/files/5773749/large.tar.gz)

Edit:
doing other tests with a 10M json file found that the json file was ISO-8859-1 encoded
converted it using 
iconv -f ISO-8859-1 -t UTF-8//TRANSLIT rmlint.json -o rmlintutf8.json
and it worked 

it seems that for some reason the resulting json file from rmlint is ISO-8859-1

SeeSpotRun:
I think it has something to do with https://github.com/sahib/rmlint/blob/master/lib/formats/json.c#L73

@sahib any thoughts on best way forwards?  Maybe write using `json-glib` instead?

sahib:
Hey,

> @sahib any thoughts on best way forwards? Maybe write using json-glib instead?

Yes, I'm afraid this is the cause. That hacky util was there before we used ``json-glib`` for replay and it didn't age well.
Alternatively, one could make [this function](https://github.com/sahib/rmlint/blob/master/lib/formats/json.c#L95) UTF-8 safe by using [g_utf8_next_char](https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html#g-utf8-next-char) and [g_utf8_get_char](https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html#g-utf8-get-char). From what I can see that's the only place that operates on individual bytes. But honestly, using ``json-glib`` is likely the more sensible option here.

SeeSpotRun:
@He1my are you able to confirm if https://github.com/SeeSpotRun/rmlint/tree/glib-json fixes this?

SeeSpotRun:
Closing for now, can re-open if issue recurs

samy-p:
Hi, I have the same error but not for the same reason :
My filenames are mostly in utf8 but I've some filenames in latin1 (old backups...)

`LC_ALL=C rmlint .

# Duplicate(s):
    ls '/tmp/test/numéros-téléphone.2002.txt'
    rm '/tmp/test/num�ros-t�l�phone.txt'`

When trying to replay :
`LC_ALL=C rmlint --replay rmlint.json .
INFO: Loading json-results `/tmp/test/rmlint.json'
[...]

Issue #463, opened 2021-01-04
Wrong detection of a json character encoding?


lockywolf:
I'm trying to deduplicate files remotely. I have run rmlint on machine A, and obtained a json file.

I uploaded it to machine B, and tried to run:
`#LC_ALL=en_US.utf8 /root/rmlint-2.10.1/rmlint --replay /mnt/SEAGATE/ // 2021-01-04_pi-rmlint-utf8.json`

I am getting an error:

```
INFO: Loading json-results `/root/2020-12-18_sorting_hdd-running-rmlint-seagate-vs-wd/pi/2021-01-04_pi-rmlint-utf8.json'
WARNING: Error: JSON data must be UTF-8 encoded
WARNING: Loading /root/2020-12-18_sorting_hdd-running-rmlint-seagate-vs-wd/pi/2021-01-04_pi-rmlint-utf8.json failed.
ERROR: No valid .json files given, aborting.
```

The json file is produced by rmlint itself, and under a utf8 locale. chardetect from python-chardet agrees that the file is utf8.

The difference between the machines is that machine A is a raspberry pi 4B, and machine B is an x86_64. rmlint versions are both 2.10.4

SeeSpotRun:
Looks like same issue as #464

SeeSpotRun:
@lockywolf are you able to confirm if https://github.com/SeeSpotRun/rmlint/tree/glib-json fixes this?

lockywolf:
I'll try to do that, although I have already crunched through the stockpile of files, and I don't have the exact test conditions.

lockywolf:
Sorry, I cannot reproduce it on my machine any more.
I must have deleted the files in question manually, or something like that.

Shall I close? If the issue reappears, somebody (maybe me) would reopen it.

SeeSpotRun:
Yes let's assume the PR fixes it until someone finds otherwise

PR #480, opened 2021-03-18
Glib json

SeeSpotRun:
Note this PR was branched off earlier PR https://github.com/sahib/rmlint/pull/479 to simplify possible merging.

Should address #464 and #463 and make json handling more robust in future.

One side-effect of this implementation is that the json file doesn't get written until right at the end.  The previous implementation writes to file as the results come in.

To do similar buffering with json-glib would require manually writing the opening and closing `[` and `]` to file, as well as the `,`'s between array elements, and using a new JsonGenerator for each element.  So it can be done but it's less elegant.  But if we we run into problems keeping the JsonArray in memory then we might have to go that way anyway.

Thoughts @sahib ?

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -515,7 +515,7 @@ AddOption(
     action='store', metavar='DIR', help='libdir name (lib or lib64)'
 )
 
-for suffix in ['libelf', 'gettext', 'fiemap', 'blkid', 'json-glib', 'gui']:
+for suffix in ['libelf', 'gettext', 'fiemap', 'blkid', 'gui']:
     AddOption(
         '--without-' + suffix, action='store_const', default=False, const=False,
         dest='with_' + suffix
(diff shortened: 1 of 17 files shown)


sahib:
Hey @SeeSpotRun,

> One side-effect of this implementation is that the json file doesn't get written until right at the end. The previous implementation writes to file as the results come in.

That's  actually a crucial property for the following reasons:

* JSON files can get very big (I saw several GB in edge cases), keeping it all in memory is not a good thing.
* Current JSON has to be written on CTRL-C, for big files that will take long.
* GUI depends on the streaming of JSON to show the results in realtime.

So, yes, I think you should go with JsonGenerator. IIRC you don't need to instance a new one on each element, but can just call ``json_generator_set_root()`` on each (or so I hope to avoid the extra allocation).

SeeSpotRun:
Yeah I kind of realised that as soon as I documented the PR.  Oh well, less elegant but not the end of the world.

Issue #458, opened 2020-12-29
--is-reflink mode uses main option parsing


Cebtenzzre:
Since the `--is-reflink` mode (documented as a "stand-alone command") uses the main option parsing code, it behaves differently from what the docs suggest.

#### Steps to Reproduce
```
$ touch a
$ touch b
$ ls rmlint.sh rmlint.json
ls: cannot access 'rmlint.sh': No such file or directory
ls: cannot access 'rmlint.json': No such file or directory

$ rmlint --is-reflink a b -o summary:stdout
$ ls rmlint.sh rmlint.json
ls: cannot access 'rmlint.sh': No such file or directory
ls: cannot access 'rmlint.json': No such file or directory

$ rmlint --is-reflink a b
$ ls rmlint.sh rmlint.json
rmlint.json  rmlint.sh
```
#### Actual Behavior
`rmlint --is-reflink` accepts the `-o` option, and creates useless output files unless `-o` is used.

#### Expected Behavior
I would expect `rmlint --is-reflink` to behave like a subcommand. It should accept only the options `-v` and `-V` as per the documentation, and should not call `rm_cmd_set_outputs` -> `rm_cmd_set_default_outputs` -> `rm_fmt_add(session->formats, "sh", "rmlint.sh")`.

It could at least be special-cased like `--dedupe` is here:
https://github.com/sahib/rmlint/blob/6abfba9af35dd89f6dc82bbfdf5957875a19db15/lib/cmdline.c#L1604-L1607

But this TODO is over three years old, maybe it's time for it to be resolved?
https://github.com/sahib/rmlint/blob/6abfba9af35dd89f6dc82bbfdf5957875a19db15/lib/cmdline.c#L1538-L1549

#### Software Versions
rmlint 2.10.1

SeeSpotRun:
Yes you are correct; it is sloppy.

SeeSpotRun:
Done and merged into https://github.com/sahib/rmlint/tree/develop.

Thanks for the input.

PR #483, opened 2021-03-26
Alt mains tidy up

SeeSpotRun:
Addresses #458

Alternate mains `--hash`, `--is-reflink` and `--dedupe` all have stand-alone option parsing.  The `--[hash|is-reflink|dedupe]` must be first arg.

Alternate main `--gui` `--shredder` does no option parsing but instead passes all args to the python script. 

Note: `rmlint --equal` still uses main option parsing but that is probable appropriate.


Cebtenzzre:
I have a feeling `rm_sys_stat` and `rm_sys_lstat` should have stayed inline because both are only one line long after preprocessing. If that's not small enough to be worth inlining then what is?

SeeSpotRun:
> I have a feeling `rm_sys_stat` and `rm_sys_lstat` should have stayed inline because both are only one line long after preprocessing. If that's not small enough to be worth inlining then what is?

You're probably right.  I had to de-inline to two procedures that called `rm_log_perror` to avoid a recursive header inclusion error.  At the same time I noticed the comment
> "We should not use functions that are I/O bound as inline functions."  ( https://www.tutorialspoint.com/when-to-use-inline-function-and-when-not-to-use-it-in-c-cplusplus )

It's probably a bit of a stretch to classify `stat` and `lstat` as I/O bound so maybe I over-reacted.

Thoughts?

Cebtenzzre:
@SeeSpotRun They are likely I/O bound in the sense that the function call overhead is negligible compared to the cost of calling `stat`. But the inline version would produce simpler machine code that might be easier for the compiler to optimize, and would avoid the extra frame in the call stack at runtime.

For single line wrapper functions, I think it's good practice to make them inline or macros regardless of what they do.

SeeSpotRun:
I was unsure of the reason for the recommendation "We should not use functions that are I/O bound as inline functions" and feared that they might make things worse.  On reading further it seems the reason is that they don't give much speed improvement, rather than that they make things worse.

> For single line wrapper functions, I think it's good practice to make them inline or macros regardless of what they do.

Ok, will revert those.

PR #485, opened 2021-04-06
Fix regex sort

### LINKED ISSUE #484: Bad regex sort when multiple paths match
Regex comparisons where both paths/basenames match incorrectly order the files, rather than allowing subsequent criteria to break the tie.

example where regex ordering breaks
1/a
1/a2
1/b
2/a
2/a2
2/b
pattern:
`x<1>r<a>l`

nloewen:
Fixes #484

### DIFF
--- a/lib/preprocess.c
+++ b/lib/preprocess.c
@@ -356,36 +356,43 @@ char *rm_pp_compile_patterns(RmSession *session, const char *sortcrit, GError **
 static int rm_pp_cmp_by_regex(GRegex *regex, int idx, RmPatternBitmask *mask_a,
                               const char *path_a, RmPatternBitmask *mask_b,
                               const char *path_b) {
-    int result = 0;
+    int result_a = 0;
+    int result_b = 0;
 
     if(RM_PATTERN_IS_CACHED(mask_a, idx)) {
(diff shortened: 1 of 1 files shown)


lib/preprocess.c
```diff
@@ -356,36 +356,43 @@ char *rm_pp_compile_patterns(RmSession *session, const char *sortcrit, GError **
...
         /* Match for the first time */
-        result = g_regex_match(regex, path_b, 0, NULL);
-        RM_PATTERN_SET_CACHED(mask_b, idx, result);
+        result_b = g_regex_match(regex, path_b, 0, NULL);
+        RM_PATTERN_SET_CACHED(mask_b, idx, result_b);
     }
 
-    if(result) {
-        return +1;
+    /* Both match */
```
SeeSpotRun, line 380: can use macro SIGN_DIFF here to simplify:
`    return SIGN_DIFF(result_b, result_a);`
Defined here: https://github.com/sahib/rmlint/blob/094fbd59cbfb2d2df73dacb0647aacc425982848/lib/utilities.h#L71

SeeSpotRun:
Thanks, beat me to it.

Issue #253, opened 2017-08-29
OS X: rmlint --gui fails with error message


workflowsguy:
I installed `rmlint 2.6.1`on 10.11.6 with `brew install rmlint`.
Running `rmlint --gui` gives

`Failed to load shredder: No module named 'shredder'
This might be due to a corrupted install; try reinstalling.`

SainteCelery:
I posted about this same problem months and months ago and was told that it must be something wrong with my environment, despite *numerous* people having had the same, seemingly unresolved, issue. Dev said he had no Mac to troubleshoot with. Still impossible to use Shredder or even see what the UI looks like.

sahib:
Hello @workflowsguy,

thanks for report. The error you're getting means that 
 Can you please provide some more info? Specifically:

- The output of ``brew install --verbose`` (to see if the GUI was actually installed)
- ``python -c 'import sys; print(sys.path)'`` 
- ``find /usr/lib -type d -iname 'shredder'``

Hello @SainteCelery,

> I posted about this same problem months and months ago and was told that it must be something wrong with my environment, despite numerous people having had the same, seemingly unresolved, issue for years. Dev said he had no Mac to troubleshoot with. No suggestions about what to look at in paths or PYTHONPATH or anything else. Reinstalled at least a dozen times. Fine-tuned paths. The issue is not my environment. And it's still impossible to use Shredder or even see what the UI looks like. The GUI should just not be listed as an option for Macs because it isn't one.

 thanks for your *nice* and *well-chosen* words, but remember that we do this for free and we do not owe you any of our time. While I somewhat understand your frustration, we're the wrong people to shout at (if there is any right person to do that) - so please calm your tone. It's hard to see for us what's happening on a platform we do not know much about and have no access to without the help from users. We know this not a nice state, but we can only properly support the platforms we're running ourselves. 

So go a few steps back, buckle up, learn about the issue and/or make someone with a Mac look at it and open a PR if you want to see this fixed as soon as possible. This is likely an issue that needs to be addressed in the homebrew formula.

workflowsguy:
Output from `brew install --verbose rmlint`:

```
==> Downloading https://homebrew.bintray.com/bottles/rmlint-2.6.1.el_capitan.bottle.tar.gz
Already downloaded: /Users/stefan/Library/Caches/Homebrew/rmlint-2.6.1.el_capitan.bottle.tar.gz
[...]

PR #491, opened 2021-04-19
clarify that mac brew install doesn't include GUI

CouldBeThis:
As discussed in sahib/rmlint#253 and Homebrew/homebrew-core#17434


SeeSpotRun:
Thanks @CouldBeThis

PR #492, opened 2021-04-19
Fix mem leaks

SeeSpotRun:


SeeSpotRun:
It's been a long while since the full testsuite ran leak-free:
```
$ sudo RM_TS_DIR=testdir RM_TS_USE_VALGRIND=1 RM_TS_CHECK_LEAKS=1 RM_TS_PEDANTIC=1 nosetests -sxv --eval-attr="not slow and not valgrind_issue"

Ran 231 tests in 8288.592s

OK
```

Issue #419, opened 2020-06-17
package the GUI as a separate binary


flokli:
Motivated by https://github.com/NixOS/nixpkgs/pull/90267.

The way `rmlint` currently starts its GUI is somewhat weird.

It writes a `.shredder-bootstrap.py.XXXXXX` file in `/tmp` containing some python code manually appending things to `sys.path`, then running

```
import shredder
shredder.run_gui()
```

This causes hard-to-debug problems for platforms where shredder isn't installed into `PYTHONPATH`, like https://github.com/sahib/rmlint/issues/253.

Also, it makes it quite complicated for packagers to move the GUI bits into a separate package output.

Instead, what about shipping `shredder` as a separate binary, with entrypoints managed via `setuptools`?

`rmlint --gui` could still be supported, by executing `shredder` from `$PATH`.

CouldBeThis:
I installed via debian `apt-get` and noticed there is one package for [`rmlint`](https://packages.debian.org/buster/rmlint), one for [`rmlint-gui`](https://packages.debian.org/buster/rmlint-gui) and another for [`rmlint-doc`](https://packages.debian.org/buster/rmlint-doc). 

A menu item is added to the GUI: `Applications > System Tools > Shredder`

Just mentioning cause this issue is still open.

SeeSpotRun:
Hey thanks @CouldBeThis and @flokli 

We dropped the `rmlint` ball a bit last year and @sahib asked me to take over maintenance at the start of this year.  I'm working my way back through open issues now and hadn't gotten back to this one.

Python / `setuptools` is not my strong suit so would be happy to see a pull request on this one.

SeeSpotRun:
This issue is now partially resolved; there is no longer a python bootstrap in /tmp, and there's a [slightly-hacky workaround](https://github.com/sahib/rmlint/blob/1a142fc0c4fd5239b99cb541863970f0c6f6bfee/gui/bin/shredder#L7-L12) in /bin/shredder to reverse-engineer `PYTHONPATH`.

I'm not planning anything further at this stage I think there's still a case for fully splitting the package so that `rmlint` installs (without gui) via
```
$ scons install rmlint
```

and `shredder` installs via
```
$ python setup.py install
```
(or possibly via `pip install`)

AutomationD:
Thanks for tremendous work! Is there any tested alpine setup? So far I got to the shredder, but can't find GtkSource. 

Here's the list of package I'm installing:
```
apk add	scons \
[...]

PR #493, opened 2021-04-20
gui: install shredder executable

SeeSpotRun:
One way of addressing #419


SeeSpotRun:
@sahib any philosophical objections to going this way?

PR #498, opened 2021-04-30
Backports

SeeSpotRun:
Catches a couple of bugfixes since release


jmigual:
Hey, when will this merge request be released? I'm experiencing the issue solved in dd4fe24 still on release 2.10.1

Cebtenzzre:
@jmigual I hear you. I personally have run into that issue, and I should probably backtrack a little and make a hotfix release based on current master and some other trivial fixes that I've accumulated. There is still an overwhelming number of problems that I haven't finished investigating or writing fixes for, but I don't think I'll reach the end of that list any time soon.

PR #499, opened 2021-05-01
--is-reflink and rm_util_link_type fixes

Cebtenzzre:
The most important change here is to allow rm_util_link_type to return RM_LINK_SYMLINK by checking for it earlier.

```
$ touch foo.txt bar.txt
$ ln -s bar.txt link.txt
$ rmlint --is-reflink foo.txt link.txt
```

Result before these changes: `Not a file` (exit status 3)
Result after these changes: `Encountered a symlink` (exit status 9)

### DIFF
--- a/docs/rmlint.1.rst
+++ b/docs/rmlint.1.rst
@@ -821,18 +821,21 @@ OTHER STAND-ALONE COMMANDS
 
 
 :``rmlint --is-reflink [-v|-V] <file1> <file2>``:
-    Tests whether ``file1`` and ``file2`` are reflinks (reference same data).
-    This command makes ``rmlint`` exit with one of the following exit codes:
+    Tests whether ``file1`` and ``file2`` are reflinked (reference same data).
+    No attempt is made to resolve symlinks. You may want to preprocess
+    arguments with ``realpath(1)`` or equivalent to resolve symlinks and for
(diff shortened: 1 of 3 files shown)


SeeSpotRun:
Thanks for contributing, much appreciated.  Feel free to add a nosetest.

SeeSpotRun:
Actually thinking about it, we should also be testing that they are links to each other, ie should expect RM_LINK_NONE for `rmlint --is-reflink foo.txt link.txt` in your example and RM_LINK_SYMLINK for `rmlint --is-reflink bar.txt link.txt`

Cebtenzzre:
Working on a nosetest. I considered testing for a symlink relationship, but rm_util_link_type is primarily for detecting reflinked files. For functions like rm_sh_emit_handler_clone, knowing that two files are hardlinked or that either is a symlink is enough to know that they aren't reflinked. And what would the link type be for a symlink to a path double, hardlink, or reflink?

A script can use `realpath` to resolve symlinks, then use the result of `rmlint --is-reflink` to identify the relationship between the destinations. IMO, the most straightforward  enhancement would be to add a `-L` flag to preprocess both arguments with `realpath(3)` - in that mode, "symlink" would not be a possible status, only "same (resolved) path".

PR #520, opened 2021-09-07
Build: allow disabling docs, do not build gschemas.compiled

sbraz:
Hi,
I am in the process of adding rmlint to Gentoo and I have a few problems, some of which I patched locally. However, I think these two should be fixed upstream.

* allow disabling docs
* do not build `gschemas.compiled`. Having this file installed causes problems for packagers. Fedora and    Arch Linux have to remove it manually as we can see here:
https://github.com/sahib/rmlint/blob/29bd07e29edf6879be933dc0e7275a90a154c00e/pkg/fedora/rmlint.spec#L48
https://github.com/archlinux/svntogit-community/blob/19a815339b610b3ee146d7e3830f3c8ac3a2a6c4/trunk/PKGBUILD#L44

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -515,7 +515,7 @@ AddOption(
     action='store', metavar='DIR', help='libdir name (lib or lib64)'
 )
 
-for suffix in ['libelf', 'gettext', 'fiemap', 'blkid', 'json-glib', 'gui']:
+for suffix in ['libelf', 'gettext', 'fiemap', 'blkid', 'json-glib', 'gui', 'docs']:
     AddOption(
         '--without-' + suffix, action='store_const', default=False, const=False,
         dest='with_' + suffix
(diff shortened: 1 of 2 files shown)


SeeSpotRun:
There have been some significant developments for package builders / maintainers in the develop branch but I fell behind in getting the next release out.

sbraz:
Hi, thanks for the reply. Actually my PR isn't perfect either as there is no way to build the manpage but not the rest of the doc. Let's wait until the next release.

By the way, I have eventually been able to add the package to Gentoo with working tests on amd64 and x86. Hopefully the next release won't require as many patches :)
https://github.com/gentoo/gentoo/commit/f87696bda2d6580aac1762fd96c5cfc95b7090ce#diff-754c644dd9146af3c5a3c306fb1a8ecd47f9dd164feac308abcbab64a0b74b66

vassilit:
We need to find a way to get the manpage included without the other docs

vassilit:
> Hi, thanks for the reply. Actually my PR isn't perfect either as there is no way to build the manpage but not the rest of the doc. Let's wait until the next release.

There is a way to do it. You can use `GetOption` in `docs/SConscript,` but as for now, HTML docs are not compiled by default, only the manpage if sphinx is found. Also, since #710, there is `--without-compile-glib-schemas` option to opt out from gschemas.compiled.

PR #523, opened 2021-09-08
Fix max sizes on x86, part of the fix for #522

### LINKED ISSUE #522: test_size.test_invalid fails on x86
Hi,
I tried to run the test suite on an x86 chroot and one test fails:
```
test_size.test_invalid ... ('--size 0-18446744073709551616',) did not trigger an error exit.
```

I was able to confirm this by running `rmlint --size 0-18446744073709551616` which indeed does not fail as excected.

sbraz:
`strtoull` returns `ULLONG_MAX,` not `ULONG_MAX`.

### DIFF
--- a/lib/cmdline.c
+++ b/lib/cmdline.c
@@ -351,7 +351,7 @@ static RmOff rm_cmd_size_string_to_bytes(const char *size_spec, GError **error) 
             return 0;
         }
 
-        if(fraction_num == ULONG_MAX && errno == ERANGE) {
+        if(fraction_num == ULLONG_MAX && errno == ERANGE) {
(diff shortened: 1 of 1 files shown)


sbraz:
Actually please don't merge this yet, `--size 0-18446744073709551615.1` doesn't trigger an error.

sbraz:
I don't really know why that fraction is supposed to fail here:
https://github.com/sahib/rmlint/blob/427791ca7c9ffa66b2c666b6b01874c5e165fae6/lib/cmdline.c#L387

I'll leave this PR in its current state, as an incomplete fix for #522.

Cebtenzzre:
Fixed in 4043c79992d251442681b72260b5425bd00e9b82.

PR #525, opened 2021-09-08
Do not compress manpage

sbraz:
Hi,
Pre-compressing the manpage causes QA warnings on Gentoo (and could also be a problem on Debian, Fedora or Arch Linux, as a fellow Gentoo developer explains [in his blog](https://cmpct.info/~sam/blog/posts/automatic-manpage-compression/)).

It would be nice to disable this feature.


sbraz:
@Cebtenzzre thanks for the other fixes, I've just pushed 2.10.2 to Gentoo :)
During the process, I stumbled on this: another minor issue that could be easily fixed for the next minor release IMO.

PR #526, opened 2021-09-09
Never hardcode compiler, select it based on CC environment variable

sbraz:
This avoids problems where `gcc`/`cc` is not the prefered compiler.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -37,8 +37,9 @@ Export('VERSION_MAJOR VERSION_MINOR VERSION_PATCH VERSION_NAME')
 def check_gcc_version(context):
     context.Message('Checking for GCC version... ')
 
+    gcc = os.environ.get("CC", "gcc")
     try:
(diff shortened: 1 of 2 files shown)


sbraz:
@Cebtenzzre please let me know if you can take a look at this before the next release. I'd be really happy if we could reduce the number of patches required to make the package work on Gentoo:
https://github.com/gentoo/gentoo/blob/1996c841ed565836c1633515175c7ecb40ec370c/app-misc/rmlint/rmlint-2.10.1-r1.ebuild#L60-L75

sbraz:
@Cebtenzzre I see that this was not merged in time for the new release so I'll have to keep the patch on Gentoo. Any chance you could look into this please?

Cebtenzzre:
Now that I've had experience getting one release out, the next one will come sooner. I can make sure this gets merged soon.

Issue #530, opened 2021-09-16
rm_offset_get_from_fd merges logically separate extents


Cebtenzzre:
### Description

When extents are physically adjacent, rmlint assumes they can safely be merged, even if they are not logically adjacent.

### Steps to reproduce

Build rmlint from the develop branch with \_RM\_OFFSET\_DEBUG=1 and fixes for #527, #528, and #529 applied.
Then run these commands:

```
$ dd if=/dev/urandom of=foo_src bs=200K oflag=sync count=1
$ xfs_io -f -c 'copy_range -l 100K foo_src' -c 'fsync' foo
$ xfs_io -f -c 'copy_range -l 100K -s 100K -d 200K foo_src' -c 'fsync' foo
$ cp --reflink foo bar
$ filefrag -vb1 foo bar
Filesystem type is: 9123683e
File size of foo is 307200 (307200 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 734768922624..734769025023: 102400:             shared
   1:   204800..  307199: 734769025024..734769127423: 102400:             last,shared,eof
foo: 1 extent found
File size of bar is 307200 (307200 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 734768922624..734769025023: 102400:             shared
   1:   204800..  307199: 734769025024..734769127423: 102400:             last,shared,eof
bar: 1 extent found
$ rmlint --is-reflink -vv foo bar
DEBUG: Testing if foo is clone of bar
DEBUG: Checking link type for foo vs bar
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=102400
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=102400
DEBUG: Offsets match at fd1=3, fd2=4, logical=0, physical=734768922624
DEBUG: Files are clones (share same data)
Link type for 'foo' and 'bar', result:
Reflink
```

### Actual result

rmlint only checks one pair of physical offsets, which ~~is probably harmless to --is-reflink but~~ indicates that rm\_offset\_get\_fiemap is merging the representation of these two extents, which is confusing and incorrect (and can cause --is-reflink false positives).

### Expected result

[...]

Issue #529, opened 2021-09-16
--is-reflink gets stuck in holes


Cebtenzzre:
### Description

rmlint --is-reflink does redundant work when met with a hole in a file.

### Steps to reproduce

Build rmlint from the develop branch with \_RM\_OFFSET\_DEBUG=1 and fixes for #527 and #528 applied.
Then run these commands:

```
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=1
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=2 seek=4
$ cp --reflink foo bar
$ filefrag -vb1 foo bar
Filesystem type is: 9123683e
File size of foo is 614400 (614400 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 699643879424..699643981823: 102400:             shared
   1:   409600..  511999: 700793454592..700793556991: 102400: 699644289024: shared
   2:   512000..  614399: 700861808640..700861911039: 102400: 700793556992: last,shared,eof
foo: 3 extents found
File size of bar is 614400 (614400 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 699643879424..699643981823: 102400:             shared
   1:   409600..  511999: 700793454592..700793556991: 102400: 699644289024: shared
   2:   512000..  614399: 700861808640..700861911039: 102400: 700793556992: last,shared,eof
bar: 3 extents found
$ rmlint --is-reflink -vv foo bar
DEBUG: Testing if foo is clone of bar
DEBUG: Checking link type for foo vs bar
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=102400
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=102400
DEBUG: Offsets match at fd1=3, fd2=4, logical=0, physical=699643879424
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=102400
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=204800
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=102400
[...]

Issue #528, opened 2021-09-16
--is-reflink detects EOF poorly


Cebtenzzre:
### Description

rmlint --is-reflink ignores the physical offset of the last extent in a file, as long as there are at least two extents total, even if a fix for #527 is applied.

### Steps to reproduce

Build rmlint from the develop branch with \_RM\_OFFSET\_DEBUG=1 and a fix for #527 applied.

For testcase 1, run these commands:

```
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=1
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=1 seek=1
$ cp --reflink foo bar
$ dd if=/dev/urandom of=bar bs=100K oflag=sync count=1 seek=1
$ filefrag -vb1 foo bar
Filesystem type is: 9123683e
File size of foo is 204800 (204800 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 694036611072..694036713471: 102400:             shared
   1:   102400..  204799: 694077128704..694077231103: 102400: 694036713472: last,eof
foo: 2 extents found
File size of bar is 204800 (204800 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 694036611072..694036713471: 102400:             shared
   1:   102400..  204799: 694103236608..694103339007: 102400: 694036713472: last,eof
bar: 2 extents found
$ cmp foo bar
foo bar differ: byte 102401, line 408
$ rmlint --is-reflink -vv foo bar
DEBUG: Testing if foo is clone of bar
DEBUG: Checking link type for foo vs bar
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=3, n_extents=1, file_offset=102400
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=0
DEBUG: rm_offset_get_fiemap: fd=4, n_extents=1, file_offset=102400
DEBUG: Offsets match at fd1=3, fd2=4, logical=0, physical=694036611072
DEBUG: Files are clones (share same data)
Link type for 'foo' and 'bar', result:
Reflink
$ echo $?
0
```

For testcase 2, run these commands:

```
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=1
$ truncate -s 200K foo
$ cp --reflink foo bar
[...]

Issue #527, opened 2021-09-16
--is-reflink skips every second extent


Cebtenzzre:
### Description

rmlint --is-reflink completely skips every second extent.

### Steps to reproduce

Build rmlint from the develop branch with \_RM\_OFFSET\_DEBUG=1.
Then run these commands:

```
$ dd if=/dev/urandom of=foo bs=100K oflag=sync count=9
$ cp --reflink foo bar
$ dd if=/dev/urandom of=bar bs=100K oflag=sync count=1 seek=1 conv=notrunc
$ filefrag -vb1 foo bar
Filesystem type is: 9123683e
File size of foo is 921600 (921600 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 672928833536..672928935935: 102400:             shared
   1:   102400..  204799: 673052336128..673052438527: 102400: 672928935936: shared
   2:   204800..  307199: 686491369472..686491471871: 102400: 673052438528: shared
   3:   307200..  409599: 690860650496..690860752895: 102400: 686491471872: shared
   4:   409600..  511999: 691562287104..691562389503: 102400: 690860752896: shared
   5:   512000..  614399: 699126407168..699126509567: 102400: 691562389504: shared
   6:   614400..  716799: 699169431552..699169533951: 102400: 699126509568: shared
   7:   716800..  819199: 699183190016..699183292415: 102400: 699169533952: shared
   8:   819200..  921599: 699530420224..699530522623: 102400: 699183292416: last,shared,eof
foo: 9 extents found
File size of bar is 921600 (921600 blocks of 1 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..  102399: 672928833536..672928935935: 102400:             shared
   1:   102400..  204799: 682882265088..682882367487: 102400: 672928935936:
   2:   204800..  307199: 686491369472..686491471871: 102400: 682882367488: shared
   3:   307200..  409599: 690860650496..690860752895: 102400: 686491471872: shared
   4:   409600..  511999: 691562287104..691562389503: 102400: 690860752896: shared
   5:   512000..  614399: 699126407168..699126509567: 102400: 691562389504: shared
   6:   614400..  716799: 699169431552..699169533951: 102400: 699126509568: shared
   7:   716800..  819199: 699183190016..699183292415: 102400: 699169533952: shared
[...]

PR #531, opened 2021-09-16
FIEMAP fixes for --is-reflink

cebtenzzre:
This pull request is intended to fix the FIEMAP-related issues with --is-reflink: #527, #528, #529, and #530.


vassilit:
(all commits of this PR have been pushed directly (without a PR) into master already)

PR #560, opened 2022-02-14
Fix a few typos and double-spacing in the man page

Silejonu:


Cebtenzzre:
It looks like since 555060fc85564333080bcd8f477b44b640b3755e there are some remaining references to `-pp`, which is no longer different from `-p`. There are a total of six in docs, this PR only changes one.

Silejonu:
> It looks like since [555060f](https://github.com/sahib/rmlint/commit/555060fc85564333080bcd8f477b44b640b3755e) there are some remaining references to `-pp`, which is no longer different from `-p`. There are a total of six in docs, this PR only changes one.

You're right, I added the remaining ones.

Silejonu:
There are a few remaining instances of -pp in the source code, but I'm not sure they should be corrected and/or how they should be:

```
CHANGELOG.md:- -pp is now silently permitted, but is the same as -p
CHANGELOG.md:* Paranoia option -pp (use -p)
CHANGELOG.md:  "paranoid" (byte-by-byte) file comparison method (option -pp), which is
tests/test_formatters/test_others.py:    for silly_option in ['-ppp', '-PPPP']:
po/rmlint.pot:msgid "Only up to -pp or down to -PP flags allowed"
po/fr.po:msgid "Only up to -pp or down to -PP flags allowed"
po/fr.po:msgstr "Seules les options de -ppp à -P sont autorisées"
po/fr.po:msgstr "Seules les options de -ppp à -P sont autorisées"
po/es.po:msgid "Only up to -pp or down to -PP flags allowed"
po/es.po:msgstr "Sólo banderas hasta -pp o menores a -PP está permitido"
po/es.po:msgstr "Sólo banderas hasta -pp o menores a -PP está permitido"
po/de.po:msgid "Only up to -pp or down to -PP flags allowed"
```

Cebtenzzre:
The changelog probably shouldn't be modified and the test is intentionally using odd yet supported paranoia flags. The .po files have old translations in them that can only really be dealt with by making `rmlint.pot` empty, running `scons xgettext` to regenerate it, then using msgattrib as described [here](https://stackoverflow.com/a/27177726) to mark the unused translations obsolete and optionally delete them, for each .po file.
But I'm not sure how the maintainers like to manage translations so it may be best to leave them alone for now.

vassilit (approved):

PR #591, opened 2022-10-16
Fix usages of single backticks to double backticks

waldyrious:
In RST, text wrapped in a \`single backtick\` is meant for "interpreted text", and is rendered using `<cite>` tags. Verbatim/literal text should instead be wrapped in \`\`double backticks\`\`, which is rendered using `<code>` tags.

See https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,8 +5,8 @@
 .. image:: https://raw.githubusercontent.com/sahib/rmlint/develop/docs/_static/logo.png
    :align: center
 
-`rmlint` finds space waste and other broken things on your filesystem and offers
-to remove it. 
+``rmlint`` finds space waste and other broken things on your filesystem and
+offers to remove it.
 
 .. image:: https://readthedocs.org/projects/rmlint/badge/?version=latest
    :target: http://rmlint.rtfd.org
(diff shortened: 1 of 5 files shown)


waldyrious:
Note: I only checked the `docs/tutorial.rst` file; it's possible that there are other files in the documentation with the same issue.

Cebtenzzre:
Based on the output of
```
grep -Prn '[^`_]`[^`_][^`]*(?<!_)`[^`_]' --include='*.rst'
```
this issue is also present in testing.rst, translators.rst, the manpage, and the README.

waldyrious:
@Cebtenzzre thanks for looking that up! Would you like write access to my fork so you could make the changes to the other files?

Cebtenzzre:
Looks like I already had access. I pushed fixes for those files, and I also corrected a few references in the manpage (e.g. `fdupes(1)`) that I think are legitimate use of interpreted text - some of them were using inline literals. How do those changes look to you?

waldyrious:
> Looks like I already had access.

Oh, right, it didn't occur to me to check that you had collaborator status in this repo. In that case, yeah, it's expected that you would have write access to the PR branch.

> I pushed fixes for those files, and I also corrected a few references in the manpage (e.g. `fdupes(1)`) that I think are legitimate use of interpreted text - some of them were using inline literals. How do those changes look to you?

Looks great! I confess I'm not 100% sure about when the interpreted text markup is supposed to be used, but those command references definitely don't look like things that should be marked up as code/verbatim.

Thanks for taking this on!

vassilit (dismissed):

vassilit (approved):

PR #602, opened 2022-12-28
tests: fix "DeprecationWarning: invalid escape sequence"

sbraz:
Before this fix, we would see
```
DeprecationWarning: invalid escape sequence '\-'
```

### DIFF
--- a/tests/test_options/test_size.py
+++ b/tests/test_options/test_size.py
@@ -61,7 +61,7 @@ def test_invalid():
             assert False
 
     # Not a valid range:
-    trigger('--size \-\-17')
+    trigger(r'--size --17')
 
     # max < min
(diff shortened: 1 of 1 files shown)


vassilit:
With Python 3.13, I don't see the warnings with nosetests but I see it with pytest :)
Don't really understand why but nevertheless, I'm not sure if `('--size --17')` was not intented in 355505edea564027e241e4a0015c9bf6bda617fd instead.

sbraz:
Just read this again. I think we should have just remove the backslashes?

vassilit:
Bonjour,

Yes I think so. That tests the range () - (-17)

vassilit (approved):

PR #603, opened 2023-01-03
Tests: replace nose with pytest

sbraz:
Hi,
nose is no longer maintained, see
https://github.com/nose-devs/nose/issues/1099#issuecomment-562394879.

This bug was initially submitted here: https://bugs.gentoo.org/878695.

This commit switches to pytest.

Amongst the changes, here are the most relevant ones:
* Create a conftest.py that contains fixtures.
* Replace nose.with_setup with yield fixtures.
* Use pytest.skip instead of print and nose.plugins.skip.SkipTest.
* Use pytest.mark.slow instead of nose.plugins.attrib.attr("slow").
* Use pytest.mark.parametrize instead of parameterized, simplify parametrized tests that use both `shell` and `inverse_order`.
* Update Travis config.
* Update SConscript.
* Update documentation.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,26 @@
 import pytest
-from tests.utils import cleanup_testdir, create_testdir
+import tests.utils as utils
 
 @pytest.fixture(autouse=True)
 def with_cleanup_between_runs():
-    cleanup_testdir()
-    create_testdir()
+    utils.cleanup_testdir()
+    utils.create_testdir()
+
+
+@pytest.fixture(params=["sh", "bash", "dash"])
+def shell(request):
+    yield request.param
+
+
+@pytest.fixture
+def usual_setup_usual_teardown():
+    utils.usual_setup_func()
+    yield
+    utils.usual_teardown_func()
+
+
+@pytest.fixture
+def usual_setup_mount_bind_teardown():
+    utils.usual_setup_func()
+    yield
+    utils.mount_bind_teardown_func()
--- a/tests/test_formatters/test_sh.py
+++ b/tests/test_formatters/test_sh.py
@@ -1,12 +1,9 @@
 #!/usr/bin/env python3
-# encoding: utf-8
+from tests.utils import *
 
 import shlex
 import subprocess
-
-from nose import with_setup
-from parameterized import parameterized
-from tests.utils import *
+import pytest
 
 
 def run_shell_script(shell, sh_path, *args):
@@ -22,9 +19,7 @@ def filter_part_of_directory(data):
     return [e for e in data if e['type'] != 'part_of_directory']
 
 
-@with_setup(usual_setup_func, usual_teardown_func)
-@parameterized([("sh", ), ("bash", ), ("dash", )])
-def test_basic(shell):
+def test_basic(usual_setup_usual_teardown, shell):
     create_file('xxx', 'a')
     create_file('xxx', 'b')
 
@@ -72,9 +67,7 @@ def test_basic(shell):
     assert '/a' in text
 
 
-@parameterized([("sh", ), ("bash", ), ("dash", )])
-@with_setup(usual_setup_func, usual_teardown_func)
-def test_paranoia(shell):
+def test_paranoia(usual_setup_usual_teardown, shell):
     create_file('xxx', 'a')
     create_file('xxx', 'b')
     create_file('xxx', 'c')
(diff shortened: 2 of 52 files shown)


sbraz:
@Cebtenzzre would it please be possible to look into this?

Cebtenzzre:
I know that nose is unmaintained, but it works fine on the distro I use (Arch Linux). I would like to replace it with something like nose2 or pytest in the future, and I will consider your changes. But it is not a priority at the moment, and I've made a lot of changes to the tests locally that would conflict with this PR.

sbraz:
@Cebtenzzre any chance you could look into this at one point? If I were to rebase this on master or develop (not sure I will because I already spent hours on this PR), would you merge it?

Cebtenzzre:
I haven't been spending much time on rmlint lately, but I would still want to push more of my own changes which include additional tests before I merge this PR.

sbraz:
Hi @cebtenzzre, FYI Gentoo is about to drop the package because it's one of the few that haven't migrated away from nose, see https://bugs.gentoo.org/878695#c5.

It saddens me as I put a lot of hours into packaging it, making the test suite work, and creating this PR. I really liked using rmlint too. I still think pytest is the way to go and nose2 is not the best solution. If you think you can fix this in the coming weeks, that would be really nice and maybe I can add it back at some point.

sbraz:
@fermino I'll try at some point, maybe it's not that much work. And I know Gentoo users miss the package.
I can't give an ETA though, it could be months before I want to dive into this. On which branch should I rebase the changes? Just to be sure I'm not missing anything?

fermino:
@sbraz Sure, no worries about it! If I happen to have some time to give it a look I will post any updates here. 

As per [this comment](https://github.com/sahib/rmlint/issues/670#issuecomment-2536566913) it should be `master`, at least for now. I'm not comfortable pushing anything to master that could pose a risk to user data (I'm barely starting to get to know the codebase), but this is only related to the tests so I think it would be a great idea in order to get the stable branch building properly in all distros. 

As a side note, you might find #677 interesting, I haven't added support for gentoo as I'm not familiar with the building process, but if we can get nose sorted out I will add it so we can keep track of its state.

a-detiste:
Same situation on Debian & derivatives for what it's worth. Only a handful of mostly unreleasable thing depending on Nose still. From users telemetry `rmlint` is top 2 just behind Debian own native toolbox `git-buildpackage` that has been fixed in git.

This GBP has a special status of "key package" (well it really ties the whole distro together). And Nose inherit transitively this property. So any day GBP author decides to upload his package will start an autoremoval counter for `rmlint`.

I don't know precisely how Ubuntu manage their fork.

vassilit:
@a-detiste, thank you for trying to keep rmlint in Debian.

Is there any way we could still succeed to be included in trixie ?

RayOei:
> Do you mind if I edit your PR to make it mergeable to master ?

This is already tied into #688 which I started looking into. But I am fine if you want to have a look as you seem to have more time available 😁

vassilit:
> This is already tied into #688 which I started looking into. But I am fine if you want to have a look as you seem to have more time available 😁

Hmm, #688 is a work-around to avoid running tests on the Python 3.12+ that introduces incompatibilities with nose. In fine, we want #603 merged to get rid of the nose dependency altogether.

RayOei:
Oops... I meant #687

sbraz:
> Do you mind if I edit your PR to make it mergeable to master ?

If you're asking me, sure, go ahead!

vassilit:
I've saved your actual commit before rebase (soon, there are a lot of conflicts) in https://github.com/sahib/rmlint/tree/603_sbraz_replace_nose.

tests/test_formatters/test_sh.py
```diff
@@ -240,9 +224,7 @@ def test_remove_empty_dirs_with_dupe_dirs(shell, inverse_order):
 
     _check_if_empty_dirs_deleted(shell, inverse_order, sh_path, data)
 
-@with_setup(usual_setup_func, usual_teardown_func)
-@parameterized([("sh", ), ("bash", ), ("dash", )])
```
vassilit, line 244: Why have you deleted the tests with sh, bash and dash ?

tests/conftest.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+from tests.utils import mount_bind_teardown_func, usual_setup_func, usual_teardown_func
+
+@pytest.fixture(params=["sh", "bash", "dash"])
```
vassilit, line 5: Ah ok, nevermind :)

vassilit:
Tests have been rebased to master, adapting the new tests that have been merged since the initial commit and functions that were moved to a different module.
The tests have been tested* on a reflink-capable FS.
@RayOei can you rebase your modifications on top of this commit ?
(*) (Should we write tests to test the test that tested those test ? :))

vassilit (approved):

PR #609, opened 2023-02-07
scons: Do not use headers with CheckFunc

fweimer-rh:
The header parameter disables the creation of fake prototypes, and CheckFunc always calls the function with no arguments, so the check fails.  This presently works by accident because compilers ignore the second #include in '#include <sys/types.h>#include <sys/xattr.h>' (so no prototype gets declared) and supply an implicit function declaration which allows calls with any number of arguments.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -195,14 +195,7 @@ def check_bigfiles(context):
         off_t_is_big_enough = False
 
     have_stat64 = True
-    if tests.CheckFunc(
-        context, 'stat64',
-        header=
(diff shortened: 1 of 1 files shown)


Cebtenzzre:
I'm aware of this issue. This is the same as PR #605 from a month ago.

PR #623, opened 2023-04-27
Add 'match-dirname' option

tomas:
Not sure if you would be interested in this. I wanted to run rmlint but only compare files among the same directory, that is, allowing duplicate files to exist in different directories.

Although this option only checks the name of the parent folder only (not the whole path in its entirety) it achieves what I needed.


tomas:
Bump! 👋

vassilit:
Does #640 meets your needs ?

I don't understand the use-case of checking the name of the parent folder only.

I think in a future evolution, we could make the filename comparaisons plugin-able.

vassilit (approved):

PR #627, opened 2023-07-31
Update .version

chapmanjacobd:
DEPRECATION: shredder 2.10.1.Ludicrous.Lemur has a non-standard version number. pip 23.3 will enforce this behaviour change. A possible replacement is to upgrade to a newer version of shredder or contact the author to suggest that they release a version with a conforming version number. Discussion can be found at https://github.com/pypa/pip/issues/12063

### DIFF
--- a/.version
+++ b/.version
@@ -1,1 +1,1 @@
-2.10.1 Ludicrous Lemur
+2.10.2


Cebtenzzre:
This was already fixed as of commit b5a6d9b359b1fc1ea75bdb6c1ae6cbdc0a304ecf.

PR #636, opened 2023-10-29
Fix example and language in manpage

mb720:
Hi and thanks a lot for rmlint!

I'd like to offer some improvements for the manpage.

It's mostly typo fixes and attempts to make the manpage easier to read.

But I think I also spotted an error in the examples. In the old version, `rmlint -e` was described as
>Find duplicates with same basename (excluding extension)

Yet I'm pretty sure that option `-b` does that, whereas `-e` makes rmlint consider two files as duplicates only if their extensions match.

### DIFF
--- a/docs/rmlint.1.rst
+++ b/docs/rmlint.1.rst
@@ -23,17 +23,17 @@ Its main focus lies on finding duplicate files and directories.
 It is able to find the following types of lint:
 
 * Duplicate files and directories (and as a by-product unique files).
-* Nonstripped Binaries (Binaries with debug symbols; needs to be explicitly enabled).
+* Nonstripped binaries (binaries with debug symbols; needs to be explicitly enabled).
 * Broken symbolic links.
 * Empty files and directories (also nested empty directories).
-* Files with broken user or group id.
+* Files with broken user or group ID.
 
 ``rmlint`` itself WILL NOT DELETE ANY FILES. It does however produce executable
 output (for example a shell script) to help you delete the files if you want
 to. Another design principle is that it should work well together with other
-tools like ``find``. Therefore we do not replicate features of other well know
+tools like ``find``. Therefore, we do not replicate features of other well-known
 programs, as for example pattern matching and finding duplicate filenames.
-However we provide many convenience options for common use cases that are hard
+However, we provide many convenience options for common use cases that are hard
 to build from scratch with standard tools.
 
 In order to find the lint, ``rmlint`` is given one or more directories to traverse.
@@ -45,13 +45,13 @@ the remaining files for duplicates.
 ``rmlint`` tries to be helpful by guessing what file of a group of duplicates
 is the **original** (i.e. the file that should not be deleted). It does this by using
 different sorting strategies that can be controlled via the ``-S`` option. By
-default it chooses the first-named path on the commandline. If two duplicates
+default it chooses the first-named path on the command line. If two duplicates
 come from the same path, it will also apply different fallback sort strategies
-(See the documentation of the ``-S`` strategy).
+(see the documentation of the ``-S`` strategy).
 
-This behaviour can be also overwritten if you know that a certain directory
+This behaviour can be also overridden if you know that a certain directory
 contains duplicates and another one originals. In this case you write the
-original directory after specifying a single ``//``  on the commandline.
+original directory after specifying a single ``//``  on the command line.
 Everything that comes after is a preferred (or a "tagged") directory. If there
 are duplicates from an unpreferred and from a preferred directory, the preferred
 one will always count as original. Special options can also be used to always
@@ -61,7 +61,7 @@ are present in both given directories (``-m``).
 We advise new users to have a short look at all options ``rmlint`` has to
 offer, and maybe test some examples before letting it run on productive data.
 WRONG ASSUMPTIONS ARE THE BIGGEST ENEMY OF YOUR DATA. There are some extended
-example at the end of this manual, but each option that is not self-explanatory
+examples at the end of this manual, but each option that is not self-explanatory
 will also try to give examples.
 
 OPTIONS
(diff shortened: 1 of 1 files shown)


docs/rmlint.1.rst
```diff
@@ -816,19 +816,19 @@ OTHER STAND-ALONE COMMANDS
...
+    Tests whether ``file1`` and ``file2`` are reflinks (they reference the same data).
     This command makes ``rmlint`` exit with one of the following exit codes:
 
     * 0: files are reflinks
     * 1: files are not reflinks
     * 3: not a regular file
     * 4: file sizes differ
     * 5: fiemaps can't be read
-    * 6: file1 and file2 are the same path
+    * 6: file1 and file2 are at the same path
```
cebtenzzre, line 827: I'm not sure about this change. This status is for when the two paths provided are literally the same path after resolving symlinks, hence no data needs to be looked at.

mb720: Maybe "path1" and "path2" are more suitable argument names. Then this could read:
>6: path1 and path2 point to the same file (after resolving symlinks)

cebtenzzre: Well, "the same file" is ambiguous here, because it could be taken to mean "the same inode." But that's a hardlink - two different paths that point to the same file. The paths themselves are equal after resolving symlinks in this case.

docs/rmlint.1.rst
```diff
@@ -723,14 +723,14 @@ FORMATTERS
 * ``stamp``:
 
   Outputs a timestamp of the time ``rmlint`` was run.
-  See also the ``--newer-than`` and ``--newer-than-stamp`` file option.
+  See also the ``--newer-than`` and ``--newer-than-stamp`` file options.
 
   Available options:
 
-  - *iso8601=[true|false]:* Write an ISO8601 formatted timestamps or seconds
+  - *iso8601=[true|false]:* Write ISO8601-formatted timestamps or seconds
```
vassilit, line 730: I guess that answers my question above ;)
I feel like it would be more accurate with a space after ISO though. What do you think?

docs/rmlint.1.rst
vassilit: Why not just explain more explicitly in the same way as we do in this discussion?
Like:
~`* 6: resolved absolute paths of file1 and file2 are identical`~
Edit: this is wrong. Paths are not and have never been resolved here.
In fact, `* 6: strcmp(path1, path2) == 0`.

docs/rmlint.1.rst
```diff
@@ -268,7 +268,7 @@ General Options
 
     The letter may also be written uppercase (similar to ``-S /
     --rank-by``) to reverse the sorting. Note that ``rmlint`` has to hold
-    back all results to the end of the run before sorting and printing.
+    back all results till the end of the run before sorting and printing.
```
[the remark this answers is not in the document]
RayOei: Not native either. But from my understanding both are fine, albeit `until` is a bit more formal and would therefore fit a bit better in manual, I think.

docs/rmlint.1.rst
```diff
@@ -374,18 +374,18 @@ Traversal Options
...
-    extension. For example: ``banana.png`` and ``Banana.jpeg`` would be considered,
+    extension. For example: ``banana.png`` and ``Banana.jpeg`` would be considered dupes,
     while ``apple.png`` and ``peach.png`` won't. The comparison is case-insensitive.
 
 :``-n --newer-than-stamp=<timestamp_filename>`` / ``-N --newer-than=<iso8601_timestamp_or_unix_timestamp>``:
 
     Only consider files (and their size siblings for duplicates) newer than a
     certain modification time (*mtime*).  The age barrier may be given as
-    seconds since the epoch or as ISO8601-Timestamp like
+    seconds since the epoch or as an ISO8601 timestamp like
```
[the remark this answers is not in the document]
RayOei: I think `8601-formatted` is unnatural in English. 
The more common phrasing would be `a timestamp formatted according to ISO 8601`

docs/rmlint.1.rst
RayOei: See above 😁

docs/rmlint.1.rst
RayOei: I find that still a bit confusing: what is the actual state it should explain?

vassilit: file1 is file2 (same path) or
file1 is a symlink to file2 or 
file2 is a symlink to file1 or 
file1 and file2 are symlink to the same file3

vassilit: But they are not symlink to each others ;)
```
vt@luga ~/dev/contrib/rmlint% ls -l b c
lrwxrwxrwx 1 vt vt 1 Mar  3 20:19 b -> c
lrwxrwxrwx 1 vt vt 1 Mar  3 20:19 c -> b
vt@luga ~/dev/contrib/rmlint% rmlint --is-reflink b c
ERROR: lib/utilities.c:1167: rm_util_link_type: Error opening /home/vt/dev/contrib/rmlint/c: Too many levels of symbolic links
```

docs/rmlint.1.rst
```diff
@@ -776,22 +776,22 @@ OTHER STAND-ALONE COMMANDS
     Start the optional graphical frontend to ``rmlint`` called ``Shredder``.
 
     This will only work when ``Shredder`` and its dependencies were installed.
-    See also: http://rmlint.readthedocs.org/en/latest/gui.html
+    See also: https://rmlint.readthedocs.io/en/latest/gui.html
 
-    The gui has its own set of options, see ``--gui --help`` for a list. These
-    should be placed at the end, ie ``rmlint --gui [options]`` when calling
-    it from commandline.
+    The GUI has its own set of options, see ``--gui --help`` for a list.  These
```
vassilit, line 781: Why two spaces after dot ?

docs/rmlint.1.rst
vassilit: @RayOei what do you think should be the right formulation ?

RayOei: I think your suggestion is fine.

docs/rmlint.1.rst
RayOei: Typo, I presume?

Issue #550, opened 2022-01-24
Filter input - only files with same relative paths


james-cook:
Is there any way to filter input to allow only files with same relative paths through.
Something like the existing **--match-basename** but for paths?

i.e. a duplicate file is only found if it shares the same relative path - with an option such as **--match-dirname**...

Pseudo code:
`rmlint <other options> --keep-all-tagged --match-dirname --match-basename DIR1 // DIR2`

(This would only work/make sense with tagging I think, and **--must-match-tagged** would not be needed in this case as files must have the same relative path in any case)

ttyusupov:
@james-cook , I've implemented something like this here: https://github.com/sahib/rmlint/pull/640

vassilit:
The case-sensitive default of `--match-relative-path` should do the work

PR #640, opened 2023-11-12
Added --match-relative-path option and support

### LINKED ISSUE #743: Is Rmlint on BSD missing duplicates?
I just ran rmlint to deduplicate two sets of folders. The first being the ones to remove duplicates from. The second being the reference directory, from which no duplicates should be found. It comes after the // between the two sets of folders.

I used rmlint . -xbegv -T "df" -km.
I then deleted all files flagged as duplicates.
Due to the -b option, this run required base names and extensions to be equivalent to flag a duplicate.

Then I ran it again without the -b option to get any files I missed on the first run
To my surprise, some files flagged as duplicates had the same basename. They also had the same extension.
I checked very carefully, putting both filenames in the same file, and carefully comparing character for character -- carefully noting any blank spaces (of which there were none) and any cases where a hyphen and an underscore might be confused. Again, none. Any cases where directional quotes might cause confusion -- or single quotes, apostrophe, or back tick. Zilch.
To my reckoning, there should be no duplicates flagged with identical names on the second run without the -b option, since all such files should have been deleted on the first run, with the -b option.
I even ran the script several times alternating with -b option and without it -- obtaining consistent results.
Am I missing something here? I've trusted this (otherwise excellent) package with deduping millions (literally) of files, and I want to make sure I understand what it's doing.
Overall, this package is a life-saver !!
Thanks in advance.

ttyusupov:
I've added `--match-relative-path` support for my own needs and seems there is an existing feature request for that: https://github.com/sahib/rmlint/issues/550.

In my case it was useful to only keep files that have been changed in different backup snapshots and keep files which have different relative paths (and different content than file with the same relative path in original copy) although the same content as some file in original copy.

### DIFF
--- a/docs/rmlint.1.rst
+++ b/docs/rmlint.1.rst
@@ -386,6 +386,12 @@ Traversal Options
     extension. For example: ``banana.png`` and ``Banana.jpeg`` would be considered dupes,
     while ``apple.png`` and ``peach.png`` won't. The comparison is case-sensitive.
 
+:``--match-relative-path``:
+
+    Only consider files as dupes that have the same path, including the filename, relative
+    to the given path. The comparison of every component is case-sensitive, unless
+    ``--case-insensitive`` is set.
+
 :``--case-insensitive``:
 
     Make the ``--match-`` options above become case-insensitive for ASCII characters
--- a/docs/rmlint.1.rst
+++ b/docs/rmlint.1.rst
@@ -363,27 +363,33 @@ Traversal Options
     they're inside duplicate directories (see ``--merge-directories``) and will
     be deleted as part of it.
 
-:``-b --match-basename``:
-
-    Only consider those files as dupes that have the same basename. See also
-    ``man 1 basename``. The comparison of the basenames is case-insensitive.
-
 :``-B --unmatched-basename``:
 
     Only consider those files as dupes that do not share the same basename.
-    See also ``man 1 basename``. The comparison of the basenames is case-insensitive.
+    See also ``man 1 basename``. The comparison is basenames case-insensitive.
+
+:``-b --match-basename``:
+
+    Only consider those files as dupes that have the same basename. See also
+    ``man 1 basename``. The comparison of the basenames is case-sensitive.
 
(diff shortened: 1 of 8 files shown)


cebtenzzre:
How are you making these backups? Typically this is something that is done with rsync (which can copy only files that have changed with --compare-dest) or with lightweight snapshots (e.g. btrfs snapshots are free because of reflinks and deleting files does not usually reduce disk usage). Making a full copy of an entire hierarchy only to delete most of it seems inefficient.

ttyusupov:
These were old backups on external drive made by TimeMachine which I needed to save somewhere else. Due to some reasons some files in snapshots were copies with exactly the same content rather than hardlinks.

Another potential scenario is when you already have a full copy on external drive but then realize you only want to keep old versions of changed files because you have a new backup on another drive.

JVAQUEROM:
Was this finally added? I am interested in this feature. However, I have the last version and cannot see it implemented.

Regards!

lib/file.c
```diff
@@ -180,6 +180,25 @@ gint rm_file_basenames_cmp(const RmFile *file_a, const RmFile *file_b) {
...
+    gint diff = file_a->depth - file_b->depth;
+    RETURN_IF_NONZERO(diff);
+    RmNode* node_a = file_a->folder;
+    RmNode* node_b = file_b->folder;
+
+    for (gint16 depth = file_a->depth; depth > 0; --depth) {
+        g_assert(node_a);
+        g_assert(node_b);
+
+        diff = g_ascii_strcasecmp(node_a->basename, node_b->basename);
```
vassilit, line 193: I know that we use _g_ascii_strcasecmp_ in `rm_file_basenames_cmp` but I think that its behaviour should be documented in the manpage as it has non-obvious consequences with today's UTF-8 filesystems (it is case-sensitive for non-ASCII characters) and with many other encodings it can create false-matches.

vassilit:
We should update the manpage and create tests for this.

vassilit (approved):

PR #663, opened 2024-05-24
scons: add missing newline for included headers

bdrung:
The header includes need to be separated by newlines. Otherwise they will end up concatenated in one line and the include will fail.

Bug-Debian: https://bugs.debian.org/1065947
Bug-Ubuntu: https://launchpad.net/bugs/2066970

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -201,9 +201,9 @@ def check_bigfiles(context):
     if tests.CheckFunc(
         context, 'stat64',
         header=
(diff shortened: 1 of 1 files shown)


vassilit:
This has been fixed with a dummy prototype in f0ca57ec907f7199e3670038d60b4702d1e1d8e2 ( #605 ).
Nevertheless, thank you for contributing

PR #668, opened 2024-05-28
scons: check *xattr functions with correct arguments

bdrung:
`SCons.Conftest.CheckFunc(context, "lgetxattr", "#include <sys/types.h>\n#include <sys/xattr.h>")` will generate following C code:

```
#include <assert.h>
#include <sys/types.h>
#include <sys/xattr.h>

#if _MSC_VER && !__INTEL_COMPILER
    #pragma function(lgetxattr)
#endif

int main(void) {
#if defined (__stub_lgetxattr) || defined (__stub___lgetxattr)
   #error "lgetxattr has a GNU stub, cannot check"
#else
   lgetxattr();
#endif

     return 0;
}
```

This code snippet will fail to compile on armhf on Ubuntu 24.04 (noble):

```
$ gcc -o test test.c
test.c: In function ‘main’:
test.c:13:11: error: too few arguments to function ‘lgetxattr’
   13 | lgetxattr();
      | ^~~~~~~~~
In file included from test2.c:3:
/usr/include/arm-linux-gnueabihf/sys/xattr.h:67:16: note: declared here
   67 | extern ssize_t lgetxattr (const char *__path, const char *__name,
      | ^~~~~~~~~
```

So pass valid arguments to the *xttr functions.

**Note**: The `funcargs` parameter requires SCons >= 4.7.0!

Bug-Ubuntu: https://launchpad.net/bugs/2066970

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -270,9 +270,14 @@ def check_posix_fadvise(context):
 def check_xattr(context):
     rc = 1
 
-    for func in ['getxattr', 'setxattr', 'removexattr', 'listxattr']:
+    for func, funcargs in (
+        ('getxattr', '"path", "name", 0, 0'),
+        ('setxattr', '"path", "name", 0, 0, 0'),
+        ('removexattr', '"path", "name"'),
+        ('listxattr', '"path", "name", 0'),
+    ):
         if tests.CheckFunc(
-            context, func,
+            context, func, funcargs=funcargs,
             header=
                 '#include <sys/types.h>'
                 '#include <sys/xattr.h>'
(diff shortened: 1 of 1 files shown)


vassilit:
```
./lib/session.c:227:(.text+0x79c): undefined reference to `rm_xattr_is_deduplicated'
/usr/bin/ld: ./lib/session.c:345:(.text+0xaf4): undefined reference to `rm_xattr_mark_deduplicated'
```

seems to be the actual error from [debian build logs](https://buildd.debian.org/status/fetch.php?pkg=rmlint&arch=armel&ver=2.9.0-2.5%2Bb1&stamp=1710031282&raw=0).

Because in [amd64](https://buildd.debian.org/status/fetch.php?pkg=rmlint&arch=amd64&ver=2.9.0-2.5%2Bb1&stamp=1710185042&raw=0):
```
Checking for C function getxattr()... (cached) yes
Checking for C function setxattr()... (cached) yes
Checking for C function removexattr()... (cached) yes
Checking for C function listxattr()... (cached) yes
Checking for C function lgetxattr()... (cached) yes
Checking for C function lsetxattr()... (cached) yes
Checking for C function lremovexattr()... (cached) yes
Checking for C function llistxattr()... (cached) yes
```

But in armel:
```
Checking for C function getxattr()... no
Checking for C function lgetxattr()... no
```

vassilit:
So the combinaison of `HAVE_FIDEDUPERANGE` or `HAVE_BTRFS_H` being true and `HAVE_XATTR` being false is not supported.

However, I think they should all be true on Ubuntu, even on armhf.

vassilit:
But [build logs](https://buildd.debian.org/status/logs.php?pkg=rmlint) are ok now.

PR #677, opened 2024-12-09
CI: add multi-distro build (using docker)

fermino:
This PR adds a github-actions job that builds rmlint on several distros, mostly to check compatibility against toolchain and libraries (and avoid stalling issues like #605). 

If there is any other distro that could be added please let me know.

### DIFF
--- a/.github/workflows/multi-distro-build.yml
+++ b/.github/workflows/multi-distro-build.yml
@@ -1,0 +1,134 @@
+name: Multi-distro build
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - master
+
+jobs:
+  multi-distro-build:
+    runs-on: ubuntu-latest
+    if: always()
+    strategy:
+      matrix:
+        build:
+          #############
+          # AlmaLinux #
+          #############
+          - {
+            tag: "almalinux:9",
+            installcmd: "dnf install -y epel-release && dnf install -y",
+            pkgs: "glib2-devel json-glib-devel gcc python3-scons",
+          }
+          - {
+            tag: "almalinux:10-kitten",
+            installcmd: "dnf install -y epel-release && dnf install -y",
+            pkgs: "glib2-devel json-glib-devel gcc python3-scons",
+          }
+          ##########
+          # Alpine #
+          ##########
+          - {
+            tag: "alpine:latest",
+            installcmd: "apk add",
+            pkgs: "build-base glib-dev json-glib-dev pkgconf scons",
+          }
+          #############
+          # ArchLinux #
+          #############
+          - {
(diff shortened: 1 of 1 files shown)


fermino:
> This is not a distro, but FreeBSD would be great.

I looked into it but right now it's [a bit convoluted](https://github.com/actions/runner/issues/385), there's this which could work though: https://github.com/marketplace/actions/freebsd-vm

RayOei (commented):
I don't know how `complete` (@fermino) you want these builds to be?
I noticed that 
* most distros (except debian & ubuntu) don't build the manpage because of `Unable to find sphinx binary in PATH`
* some warnings for scons (except Debian & Ubuntu): 
`scons: Reading SConscript files ... 
/build/SConstruct:831: SyntaxWarning: invalid escape sequence '\.'
  'sed -i "s/2\.0\.0/{v}/g" po/rmlint.pot',
/build/SConstruct:832: SyntaxWarning: invalid escape sequence '\('
  'sed -i "s/^Version:\(\s*\)2\.0\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'`
* Use `apt-get` instead of `apt` in the `Debian` section - this will not trigger the apt warning in the build
* Some complain about: '/bin/sh: line 1: git: command not found
Checking for git revision... Unable to find git revision.`

fermino:
@RayOei I mostly wanted to make sure that things are building properly as some things have been breaking in arch and gentoo (and IIRC there is some soon to be packaging issues in debian).

Thanks for looking into those messages, I will fix what I can and get back with whatever's left :)

vassilit:
Something useful would be to build on macOS (intel, arm) and also on 32bit architectures.

RayOei:
> Something useful would be to build on macOS (intel, arm) and also on 32bit architectures.

Do we want to add those first before moving on with this PR?
I think most of the build errors should be solved by other commits by now.
I would also want to see how we detect errors & warnings a bit better in the workflow, so things don't get overlooked.

vassilit:
I've rebased to master just to check if we still compile correctly on all distributions.

I don't have an opinion if you should add macOS in this PR or another.

fermino:
Hey there, I'm not having the time right now to work on it, but besides that I think CirrusCI is a good option if we want to automate things in FreeBSD, and also AFAIR there is some support in github actions for mac, not sure if ARM only or x86 too. In any case, I'd rather have something merged and working than nothing at all™ :laughing:

fermino:
In the end, building ok does not guarantee correctness, so it might be a good idea to also test the whole thing in one FreeBSD/Darwin/whatever-other-unix and then have compilation-only checks for every distro/release.

vassilit (approved):

PR #678, opened 2024-12-16
docs: add repology badge to readme

fermino:


vassilit:
Is there a way to avoid rebuilding and running the whole testsuite for changes that are outside the test scope ?
To save minutes for both saving quotas and the environment :)

RayOei:
> Is there a way to avoid rebuilding and running the whole testsuite for changes that are outside the test scope ? To save minutes for both saving quotas and the environment :)

In principal: yes. I haven't looked into that yet, here. But in my daily shop I am accustomed to have the build be smart enough to not redo the work if there is nothing to do (and with builds running for hours that matters a lot).
I will add it as an `todo` with the testing.

vassilit:
Not sure what to think about the list of maintainers listed in https://repology.org/project/rmlint/information

vassilit:
Debian will need to remove this link while packaging for privacy reason.

PR #683, opened 2025-02-28
Fix #679 Illegal number

### LINKED ISSUE #679: "illegal number" warning running script generated with --merge-directories (-D)
I see a shell error message when running rmlint.sh generated by a command like `rmlint --merge-directories A B` (aka -D option).

With dash, the error message is `./rmlint.sh: 161: [: Illegal number:` with `0` printed a few lines later. With bash, the error message is `./rmlint.sh: line 161: [: ` and then a few lines later `0: integer expression expected`.

Steps to reproduce:
```
mkdir -p /tmp/rmlint-bug/dir
cd /tmp/rmlint-bug
echo "test" > dir/file
cp -r dir dir2
rmlint --merge-directories dir dir2
./rmlint.sh -p -d
```

On my system, `/bin/sh` is a symlink to dash. `dpkg -s dash` says "Version: 0.5.12-2". Output from the above is:

```
$ mkdir -p /tmp/rmlint-bug/dir
$ cd /tmp/rmlint-bug
$ echo "test" > dir/file
$ cp -r dir dir2
$ rmlint --merge-directories dir dir2

# Duplicate Directorie(s):
    ls -la '/tmp/rmlint-bug/dir'
    rm -rf '/tmp/rmlint-bug/dir2'

==> Note: Please use the saved script below for removal, not the above output.
==> In total 2 files, whereof 1 are duplicates in 1 groups.
==> This equals 5 B of duplicates which could be removed.
==> Scanning took in total 0.061s.

Wrote a json file to: /tmp/rmlint-bug/rmlint.json
Wrote a sh file to: /tmp/rmlint-bug/rmlint.sh
$ ./rmlint.sh -p -d
# ///This script will be deleted after it runs///
[  0%] Keeping:  /tmp/rmlint-bug/dir
[ 50%] Deleting: /tmp/rmlint-bug/dir2
./rmlint.sh: 161: [: Illegal number: 
# Duplicate Directorie(s):
    ls -la '/tmp/rmlint-bug/dir'
    rm -rf '/tmp/rmlint-bug/dir2'

==> Note: Please use the saved script below for removal, not the above output.
==> In total 2 files, whereof 1 are duplicates in 1 groups.
==> This equals 5 B of duplicates which could be removed.
==> Scanning took in total 0.056s.

Wrote a json file to: /tmp/rmlint-bug/rmlint.json
Wrote a sh file to: /tmp/rmlint-bug/rmlint.sh
0
[100%] Done!
Deleting script  ./rmlint.sh
```

If I change the shebang in rmlint.sh to `#!/bin/bash` before running it, the output is slightly different. 

```
...
$ ./rmlint.sh -p -d
# ///This script will be deleted after it runs///
[  0%] Keeping:  /tmp/rmlint-bug/dir
[ 50%] Deleting: /tmp/rmlint-bug/dir2
./rmlint.sh: line 161: [: 
# Duplicate Directorie(s):
    ls -la '/tmp/rmlint-bug/dir'
    rm -rf '/tmp/rmlint-bug/dir2'

==> Note: Please use the saved script below for removal, not the above output.
==> In total 2 files, whereof 1 are duplicates in 1 groups.
==> This equals 5 B of duplicates which could be removed.
==> Scanning took in total 0.060s.

Wrote a json file to: /tmp/rmlint-bug/rmlint.json
Wrote a sh file to: /tmp/rmlint-bug/rmlint.sh
0: integer expression expected
[100%] Done!
Deleting script  ./rmlint.sh
```

`bash --version` outputs "GNU bash, version 5.2.15(1)-release (x86_64-pc-linux-gnu)".

I checked that if I run `echo "different" > dir2/file` after rmlint but before rmlint.sh, the script correctly refuses to delete `dir2`, so it seems that the behavior is correct despite the error. Assuming this holds for more complicated situations, this bug is rather low-priority. However, given that I'm using the "paranoid" option, this error makes me feel even more paranoid. 😉

RayOei:
Solves "Illegal number" #679 error when two _directories_ are rechecked when generated `rmlint.sh` is executed with the `-p` parameter before removing.

### DIFF
--- a/lib/formats/sh.sh
+++ b/lib/formats/sh.sh
@@ -130,11 +130,9 @@ check_for_equality() {
     if [ -f "$1" ]; then
         # Use the more lightweight builtin `cmp` for regular files:
         cmp -s "$1" "$2"
-        echo $?
     else
         # Fallback to `rmlint --equal` for directories:
         "$RMLINT_BINARY" -p --equal %s "$1" "$2"
-        echo $?
     fi
 }
 
@@ -159,7 +157,8 @@ original_check() {
     if [ -z "$DO_PARANOID_CHECK" ]; then
         return 0
     else
-        if [ "$(check_for_equality "$1" "$2")" -ne "0" ]; then
+        # Check latest result.
+        if ! check_for_equality "$1" "$2"; then
             echo "${COL_RED}^^^^^^ Error: files no longer identical - cancelling.....${COL_RESET}"
             return 1
         fi


lib/formats/sh.sh
```diff
@@ -159,7 +157,10 @@ original_check() {
     if [ -z "$DO_PARANOID_CHECK" ]; then
         return 0
     else
-        if [ "$(check_for_equality "$1" "$2")" -ne "0" ]; then
+        # Run check
+        check_for_equality "$1" "$2"; check_return = $?
+        # Check latest result. Note the directory compare returns text, which would fail this function #679
+        if [ "$check_return" -ne "0" ]; then
```
vassilit, line 163: What about  `if ! check_for_equality "$1" "$2"; then` ?

RayOei: That _is_ shorter. 
I am not a bash guru 😁 but the first is a bit more clear _to me_ that we are checking the `return value` of the function as the function itself is returning the full stdout of the compare - which is the reason the `if` failed to begin with as the comparison wasn't comparing the things that were expected.
Result-wise they are the same. Having said that: I did make a mistake when adding and pushing the `check_return` 😊 It should read `check_return=$?`. So this needs an update, either way.

vassilit: Do not mention _bash_ :) 
This should not be a bash-script but an _sh_ POSIX-compatible script.
Well, it is shorter and it is what b2dc4acc49b11726a7504d8a88d065f646aa4b01 does, which will eventually get merged.
So I think we should use that.

I propose that your PR be merged for 2.10.3 and it will be overwritten by b2dc4acc49b11726a7504d8a88d065f646aa4b01 when merging for 2.11.0

RayOei: You make a lot of ground 😁 
It seems we are patching stuff that is already patched. Ah well... I fixed my my fix.

RayOei: Not directly related but with testing it bites me that the version has not been updated (e.g. testing with the incorrect executable). Mind if we bump the version to 2.10.3 so it becomes clear what we are working on?
Can make a separate PR for that for clarity, obviously.

vassilit: > It seems we are patching stuff that is already patched.

Yes, the idea was to use specifically `if ! check_for_equality "$1" "$2"; then` in your PR so that line will not conflict when merging later.

vassilit: > Not directly related but with testing it bites me that the version has not been updated (e.g. testing with the incorrect executable).

Do you get a warning of some sort ?

RayOei: > line will not conflict when merging later.

Ah.. that makes sense 👍 

> Do you get a warning of some sort ?

Nothing to worry about, just me _not_ working with the exec I just build (and therefore getting confusing results)

vassilit: > Nothing to worry about, just me not working with the exec I just build (and therefore getting confusing results)

When you run `RM_TS_PRINT_CMD=1  RM_TS_KEEP_TESTDIR=1 nosetests -s tests/test_formatters/test_sh.py` to test this PR, it compiles and run rmlint from the repository, not from the system path

RayOei: 😁 I was not running the tests but trying to replay an issue report with something I modified to see whether my idea was sound. `which` told me enough but I was still second-guessing myself. Either way... having a new version after a release is always sound in my book. The current develop/master issue is adding some confusion too, which makes me a bit more cautious than if I were free hacking away on develop. Thanks for #694

vassilit (approved):

PR #684, opened 2025-02-28
SCons: fix release Python warnings

### LINKED ISSUE #655: SyntaxWarning: invalid escape sequence
If scons uses python 3.12 or later, these warnings appear:

```
scons: Reading SConscript files ...
SConstruct:839: SyntaxWarning: invalid escape sequence '\.'
  'sed -i "s/2\.0\.0/{v}/g" po/rmlint.pot',
SConstruct:840: SyntaxWarning: invalid escape sequence '\('
  'sed -i "s/^Version:\(\s*\)2\.0\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'
```

See https://docs.python.org/3/library/re.html:

> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with `'r'`. So `r"\n"` is a two-character string containing `'\'` and `'n'`, while `"\n"` is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

vassilit:
Closes #655

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -828,8 +828,8 @@ if 'release' in COMMAND_LINE_TARGETS:
         )
 
         cmds = [
-            'sed -i "s/2\.0\.0/{v}/g" po/rmlint.pot',
-            'sed -i "s/^Version:\(\s*\)2\.0\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'
+            r'sed -i "s/2\.8\.0/{v}/g" po/rmlint.pot',
+            r'sed -i "s/^Version:\(\s*\)2\.8\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'
         ]
 
         for cmd in cmds:


RayOei:
This would also improve PR #677

SConstruct
```diff
@@ -828,8 +828,8 @@ if 'release' in COMMAND_LINE_TARGETS:
         )
 
         cmds = [
-            'sed -i "s/2\.0\.0/{v}/g" po/rmlint.pot',
-            'sed -i "s/^Version:\(\s*\)2\.0\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'
+            r'sed -i "s/2\.0\.0/{v}/g" po/rmlint.pot',
+            r'sed -i "s/^Version:\(\s*\)2\.0\.0/Version:\\1{v}/g" pkg/fedora/rmlint.spec'
```
fermino, line 832: The first line looks okay to me. I think the second one needs un-escaping the parentheses and the \1 reference. I'd like to have a test case for building in fedora, although I think this would be for a release with python 2...

fermino: Nevermind, I totally misread it. For what I understand it replaces the version string in a translation and a packaging file. It is actually broken as the replaced string is 2.0.0 and the files are already in 2.8.0. 

@sahib I'm guessing that was intended to be used locally when releasing tarballs in github?

fermino: That looks great! I would personally prefer an automated release process with CI, but in the meantime fixing the script would be awesome!

RayOei: Looks fine. I was waiting for @fermino to accept. Will check back in a few days 🕚

fermino:
Hey guys, I'm just back from a couple of vacation days and a (hopefully not bad) wrist accident. I saw tons of activity in the repo! I will try to catch up with all the emails but in the meantime: where is help/approvals needed?

fermino:
Regarding the PR: looks good to me! I don't want to merge it right now as I'm not sure if there's anything that you guys want merged before, but to me it's good to go.

fermino (approved):

RayOei:
> a (hopefully not bad) wrist accident

Ouch... take care!!

Issue #685, opened 2025-03-01
Checking for C11 generates an empty '-.o' file


vassilit:
```
[...]
Checking for -std=c11 support...
[...]
execve("/usr/lib/gcc/x86_64-pc-linux-gnu/14.2.1/cc1", ["/usr/lib/gcc/x86_64-pc-linux-gnu"..., "-quiet", "-", "-quiet", "-d
umpbase", "-", "-mtune=generic", "-march=x86-64", "-std=c11", "-o", "/tmp/ccO8WMyi.s"], 0x38f24630 /* 78 vars */) = 0
[...]
execve("/usr/bin/as", ["as", "--64", "-o", "-.o", "/tmp/ccO8WMyi.s"], 0x38f24630 /* 78 vars */) = 0
```

PR #686, opened 2025-03-01
SCons: avoid checks leftovers

vassilit:
Closes: #685

`echo "" | cc -xc - -c` generates a '-.o' leftover (gcc/clang).

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       RM_TS_DIR: "/tmp/rmlint-unit-testdir"
     steps:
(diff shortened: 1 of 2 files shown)


fermino (approved):
Thanks for the fix, I will merge it as soon as it passes CI.

fermino:
@vassilit would you mind changing https://github.com/actions/runner-images/issues/10636 to `ubuntu-22.04`? ubuntu-latest is now 24.04 and there has python imp's module seems to have been removed. 

This change should eventually be caught by #677, but for now it'd be nice to have this merged. 
If you add it in a different commit I will fast-forward master so the changes are not mixed. 

Thanks again! :)

Related issue: https://github.com/actions/runner-images/issues/10636

vassilit:
I merged this PR, because this was marked as approved by you, but doing so I voided your PR #688, without intending to do so. Sorry :)

Issue #687, opened 2025-03-01
Test suite: replace nose with pytest [2025] (#603 follow-up)


fermino:
This is a follow-up on the great work by sbraz on #603.

The objective is to migrate away from nose to pytest and given that quite some things have changed in master  I'm thinking on splitting the work on a per-file basis. In the meantime things will be compatible and it will be a lot easier to review and approve changes in small batches. I will keep the list updated with referenced PRs as we get through it. 

Any help is appreciated! :)

Test files: 
- [x] tests/requirements.txt
- [x] tests/SConscript
- [x] tests/test_formatters/test_csv.py
- [x] tests/test_formatters/test_json.py
- [x] tests/test_formatters/test_others.py
- [x] tests/test_formatters/test_py.py
- [ ] tests/test_formatters/test_sh.py
- [ ] tests/test_mains/test_dedupe.py
- [ ] tests/test_mains/test_hash.py
- [ ] tests/test_mains/test_is_reflink.py
- [ ] tests/test_options/test_cache.py
- [ ] tests/test_options/test_clamp.py
- [ ] tests/test_options/test_color.py
- [ ] tests/test_options/test_equal.py
- [ ] tests/test_options/test_help.py
- [ ] tests/test_options/test_hidden_dirs.py
- [ ] tests/test_options/test_keep_hardlinks.py
- [ ] tests/test_options/test_km.py
- [ ] tests/test_options/test_match_basenames.py
- [ ] tests/test_options/test_match_with_extension.py
- [ ] tests/test_options/test_match_without_extension.py
- [ ] tests/test_options/test_merge_directories.py
- [ ] tests/test_options/test_mtime_window.py
- [ ] tests/test_options/test_newer_than.py
- [ ] tests/test_options/test_no_backup.py
- [ ] tests/test_options/test_pass_files.py
- [ ] tests/test_options/test_perms.py
- [ ] tests/test_options/test_rankby.py
- [ ] tests/test_options/test_replay.py
- [ ] tests/test_options/test_size.py
- [ ] tests/test_options/test_sorting.py
- [ ] tests/test_options/test_stdin.py
- [ ] tests/test_options/test_symlinks.py
- [ ] tests/test_robustness/test_8Mlongpathfiles.py
- [ ] tests/test_robustness/test_badlinks_as_args.py
- [ ] tests/test_robustness/test_bigfiles.py
- [ ] tests/test_robustness/test_collisions.py
- [ ] tests/test_robustness/test_manyfiles.py
[...]

PR #688, opened 2025-03-01
ci: use ubuntu-22.04 to run tests

fermino:
There have been some changes in ubuntu24 that make it incompatible with nose. This will be eventually fixed but for now it will be good to have the test suite running.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     env:
       RM_TS_DIR: "/tmp/rmlint-unit-testdir"
     steps:


fermino:
@sahib would you mind disabling the "approval" requirement on MRs at least until we have sufficient activity and reviewers? I'll make sure to discuss non-trivial changes with other people but it will be nice to be able to have some changes pushed without that. (This change is needed as it's blocking a couple PRs). 
Thank you!

sahib:
@fermino Sure, I disabled it for now.

By the way: I get notifications when I get tagged explicitly. Best way to get me do any admin stuff would be to @sahib therefore.

RayOei:
> Keep in mind that this lets people merge their own PRs though.

@fermino @vassilit Ok with reinstating the PR review check?

vassilit:
> > Keep in mind that this lets people merge their own PRs though.
> @fermino @vassilit Ok with reinstating the PR review check?

Yes, certainly

vassilit:
> There have been some changes in ubuntu24 that make it incompatible with nose. This will be eventually fixed 

I think they will never get fixed upstream.
Nose is unmaintained and use modules incompatibles with Python 3.12+. Debian used to have patch for that, but I see nose is not anymore in testing nor in sid. As Ubuntu is based on it, they won't restore nose-compatibility.
We are installing nose from pip rather than via packaging, so as a workaround what we could do for now if we want -latest is to patch it like as distributions do (https://github.com/sahib/rmlint/pull/688#issuecomment-2692630618) and it should work.

RayOei:
> Keep in mind that this lets people merge their own PRs though.

@sahib You can revert that permission now. Thanks!

RayOei:
> This will be eventually fixed
I _think_ he meant by rmlint, not by Ubuntu 😁 
Issue #687 should resolve that as it will remove all dependencies. I am looking into that, based on the work done in #603. There are still a few nose references but by the looks of it, it should not be too hard to remove those as well.

RayOei:
I think this one can be closed (not merged) as #687 should provide a more future proof solution.

Issue #672, opened 2024-09-15
Data loss with --replay and hard-linked files due to missing checksums.


misieck:
Under certain conditions, rmlint does not calculate checksums of encountered files. **All entries without checksums are seemingly interpreted by --replay as one and the same file, irrelevant of their content. This results in deletion of all but one of such files.** 
One such condition that I found is when encountering hard links. Consider the following test setup:

```
\testdir
    - file_one 
    - file_one_link
    - file_two 
    - file_two_link
```

Sizes of file_one and file_two must differ. Running rmlint produces a rmlint.json file with no checksums in it, but the actions are correct.

```
# Duplicate(s):
    ls '/home/muser/rmlint/test/two'
    rm '/home/muser/rmlint/test/two_link'
    ls '/home/muser/rmlint/test/one'
    rm '/home/muser/rmlint/test/one_link'

```
Running rmlint --replay rmlint.json afterwards produces a script attempting to delete all but one of the listed files:
```
# Duplicate(s):
    ls '/home/muser/rmlint/test/one'
    rm '/home/muser/rmlint/test/one_link'
    rm '/home/muser/rmlint/test/two'
    rm '/home/muser/rmlint/test/two_link'
```

This would delete both two and two_link.

vassilit:
I reproduce with `version 2.10.2 compiled: Jan  3 2025 at [14:24:47] "Ludicrous Lemur" (rev f0ca57ec)`

StatusCode404:
It is in the man page for "--replay", well I clearly states that it won't do the default  "--hardlinked"...
```
By design, some options will not have any effect. Those are:

    --followlinks
    --algorithm
    --paranoid
    --clamp-low
    --hardlinked
    --write-unfinished
    … and all other caching options below.
```
https://rmlint.readthedocs.io/en/master/rmlint.1.html

StatusCode404:
This is not a bug since it is by design.
This should go under a feature request for the next version.

misieck:
No,  --hardlinked, means report hard links as duplicates. The negative of that is not to report them as duplicates and thus preserve both files.
 *rmlint deletes both files*. There is no world this could be by design. People are losing data with this.

vassilit:
Indeed, if rmlint would only delete all but one reference to the same inode, the only lost information would be the file entry in those directories, so basically the name, some attributes and path. This is by design.

Here rmlint deletes all references to the same inode, so you lose the content of the file(s).

PR #689, opened 2025-03-02
fix data-loss issue with replays w/o checksums

vassilit:
Closes: #672

See bug details in https://github.com/sahib/rmlint/issues/672#issuecomment-2692483380.

This is just a bugfix, but I feel we could fix a bit more, although I'm not sure if that wouldn't introduce a regression. I haven't checked all the consumers of `file->digest`.
 
In `replay.c` we do:
```c
    file->digest = rm_digest_new(RM_DIGEST_EXT, 0);
[...]
    /* Fake the checksum using RM_DIGEST_EXT */
    JsonNode *cksum_node = json_object_get_member(object, "checksum");
    if(cksum_node != NULL) {
        const char *cksum = json_object_get_string_member(object, "checksum");
        if(cksum != NULL) {
            rm_digest_update(file->digest, (unsigned char *)cksum, strlen(cksum));
        }
    }
```

I would do instead:

```c
    file->digest = NULL;
[...]
    /* Fake the checksum using RM_DIGEST_EXT */
    JsonNode *cksum_node = json_object_get_member(object, "checksum");
    if(cksum_node != NULL) {
        const char *cksum = json_object_get_string_member(object, "checksum");
        if(cksum != NULL) {
            file->digest = rm_digest_new(RM_DIGEST_EXT, 0);
            rm_digest_update(file->digest, (unsigned char *)cksum, strlen(cksum));
        }
    }
```
And reintroduce checks for `file->digest == NULL` ; that would avoid pointless allocations.
Another way to be safe would be to check for `file->inode` to confirm that they belong to the same group.
Are hardlinks the only situation when rmlint does not generate checksums ?

PS: @sahib if you still follow this repository, the bird theme is nice. It's a change from the overly formal naming in code of other projects :)

### DIFF
--- a/lib/checksum.c
+++ b/lib/checksum.c
@@ -991,6 +991,10 @@ gboolean rm_digest_equal(RmDigest *a, RmDigest *b) {
         return false;
     }
 
+    if(a->bytes == 0) {
+        return true; /* this non-sense is used in replays */
+    }
+
     if(a->type == RM_DIGEST_PARANOID) {
         RmParanoid *pa = a->state;
         RmParanoid *pb = b->state;
--- a/lib/replay.c
+++ b/lib/replay.c
@@ -834,7 +834,7 @@ bool rm_parrot_cage_load(RmParrotCage *cage, const char *json_path, bool is_pref
 
         rm_log_debug("[okay]\n");
 
-        if(file->digest != NULL && !rm_digest_equal(file->digest, last_digest)) {
+        if((!file->digest->bytes && file->is_original) || !rm_digest_equal(file->digest, last_digest)) {
             rm_digest_free(last_digest);
             last_digest = rm_digest_copy(file->digest);
             rm_parrot_cage_push_to_group(cage, &group, false);
@@ -873,6 +873,9 @@ static void rm_parrot_merge_identical_groups(RmParrotCage *cage) {
     for(GList *iter = cage->groups->head; iter; iter = iter->next) {
         GQueue *group = iter->data;
         RmFile *head_file = group->head->data;
+
+        if (!head_file->digest->bytes)
+            continue;
 
         GQueue *existing_group = g_hash_table_lookup(digest_to_group, head_file->digest);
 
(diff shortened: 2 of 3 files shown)


RayOei:
I would suggest to move this to its own issue as a future improvement, so it doesn't get forgotten?

vassilit:
> I would suggest to move this to its own issue as a future improvement, so it doesn't get forgotten?

I don't understand what you mean by that

vassilit:
Ah, you mean the superfluous allocation. Got it.

vassilit:
> @vassilit I haven't had the time, yet, to dive into this. Are there, besides the case described in #672, other scenarios you would want to have checked, looking at the tag?

Just that it does not do silly thing, but as it passes all tests, I guess we're safe.

RayOei (approved):
I have tried several extended cases. All were fine 👍 

I did bump into an issue with symbolic/soft links but can't find tests to explain how the behavior is expected to be.
If you take the error case and use soft instead of hard link the duplicate file is removed but the soft link is not. If you have many of those you are left with a lot dead links. That doesn't seem expected.

I note it here but I assume it needs its own issue? Unless this is intended?

vassilit:
> If you take the error case and use soft instead of hard link the duplicate file is removed but the soft link is not. 

I will test it a bit later, in the meantime I do not merge this fix.

vassilit:
> I did bump into an issue with symbolic/soft links but can't find tests to explain how the behavior is expected to be. If you take the error case and use soft instead of hard link the duplicate file is removed but the soft link is not. If you have many of those you are left with a lot dead links. That doesn't seem expected.

Can you details the commands that you used to generate this case ? I'm not sure I understand.

fermino:
Would a subsequent pass detect the broken links?

RayOei:
It does.. but that was a bit the base of my question. Should you expect to run it again to do some additional cleaning or would it be more logical that the first clean _also_ cleans the soft links. I couldn't find conclusive information on how this is supposed to be treated. It is not a _big_ issue but I can image that some users after running the generated script assume things have been cleared.

vassilit:
> It does.. but that was a bit the base of my question. Should you expect to run it again to do some additional cleaning or would it be more logical that the first clean _also_ cleans the soft links? 

That's an interesting point, but unrelated to this bug that concerns replays exclusively.

We could do two passes and check for dangling links but I'm not sure if that would be a bug fix or a feature.

RayOei:
>  I'm not sure if that would be a bug fix or a feature

Me neither :grin
But personally I would (like to) see that as an improvement, either way. I will create an issue with the information above.

Issue #682, opened 2025-02-11
rmlint writes non-compliant JSON if quotes are passed at the command line


RollingStar:
If you set `-o json` and your command contains `"`, the json will look like

```
"args": "rmlint path1 path2 --rank-by=mplaO --no-followlinks --keep-hardlinked -o json:asdf.json -o pretty:stdout -o summary:stdout -o sh:asdf.sh -c sh:cmd=.asdf.sh "$1" "$2"",
```

`$2"",`

is not compliant - the " is not escaped. JSON parsers fail on this.

PR #691, opened 2025-03-02
json: properly escape cwd and args

vassilit:
as they may contain any char but \0.

Closes: #682


vassilit:
This is better fixed in `develop`, by using JSON-Glib (but `--without-json-glib` _may_ have to disappear), but for 2.10.3, this bug fix is enough.

Related: #480

RayOei (approved):
👍

PR #695, opened 2025-03-03
chat

vassilit:


vassilit:
> Hey there! :) Is there people in the IRC channel? If there is I could try to spend some time there. If there's not much activity it _might_ be better to try to funnel down people here instead. Any thoughts?

There is «sometimes» people :) I would say very rarely.
But IRC sometimes is a good way to coordinate things without getting into long threads.

I'd say it's complementary. I see Issue as a bug/feature-tracker, not a mailing-list nor a chat-channel.

RayOei:
I never am, anymore. That was ages ago.
I think that is why GH has `Discussions` to keep everything in the open for everybody but it is not configured here. YMMV.

vassilit:
So do you think we should delete this line altogether ?

RayOei:
Well... if somebody _is_ present there, albeit seldom, it would still be fine.
We can also ask sahib to activate `Discussions` and remove the `irc` reference?

RayOei:
And in spirit of [keep communication public](https://opensource.guide/best-practices/#keep-communication-public) using `discussion`, now-a-days may be the better choice.

fermino:
Although I love IRC I do think it doesn't make much sense to keep it. @sahib would you mind enabling the discussions feature in the repo?

vassilit:
> Although I love IRC I do think it doesn't make much sense to keep it. @sahib would you mind enabling the discussions feature in the repo?

The last commit delete the reference to IRC altogether

fermino (approved):

fermino:
Yeah, I saw it, just wanted to chime in :)

PR #701, opened 2025-03-05
various C linting

vassilit:
- add a check for a truncated path (PATH_MAX is indicative only, the real maximum length of a path is file-system dependent)
- remove an unused variable
- add some static and const
- simplify header path
- add IWYU to scons


RayOei (approved):
👍

PR #706, opened 2025-03-08
git: ignore tests-venv and py-format generated JSON

vassilit:


RayOei (approved):
Top 👍 
I had also added most of those locally already 😁

PR #707, opened 2025-03-09
set the minimum Glib version to 2.64

vassilit:


vassilit:
@RayOei I even wonder if we should not release this as well for 2.10.3. 
Having different Glib requirement for 32-bit and 64-bit users is a bit strange. 

With 2.64, both 32 and 64 bit works well.

vassilit:
This will better reflect the fact that 64-bit version already, including 2.10.2, requires glibc 2.64.

RayOei:
> release this as well for 2.10.3.

Be my guest.

vassilit:
@RayOei  The warning should have disappeared.

PR #708, opened 2025-03-09
remove GLib deprecation warnings

vassilit:


fermino:
What is the reasoning behind the max version? (I mean, do we have any features that will fail with a newer one?)

vassilit:
> What is the reasoning behind the max version? (I mean, do we have any features that will fail with a newer one?)

There are multiples depreciation warnings, see https://gitlab.gnome.org/GNOME/glib/-/commit/aac56f1618aabfcf4c6b3ef1ee5b87322208e9ad

fermino:
> There are multiples depreciation warnings, see https://gitlab.gnome.org/GNOME/glib/-/commit/aac56f1618aabfcf4c6b3ef1ee5b87322208e9ad

Awesome, I just saw #707. Approved.

fermino (approved):

vassilit:
Just to be clear, `GLIB_VERSION_MAX_ALLOWED` does not mean at all you cannot compile/link/use a recent version of GLib.
It's just that the macros in GLib will make sure that the API they present to rmlint is stable.

PR #709, opened 2025-03-10
shredder: use compile-time length

vassilit:


fermino:
Looks good to me! Do you want this merged asap or should we wait for something else? (I haven't fully understood what the idea about master/develop is so I don't want to mess anything up :p)

lib/cmdline.c
```diff
@@ -176,7 +176,7 @@ static void rm_cmd_start_gui(int argc, const char **argv) {
         return;
     }
 
-    if(write(bootstrap_fd, RM_PY_BOOTSTRAP, strlen(RM_PY_BOOTSTRAP)) < 0) {
+    if(write(bootstrap_fd, rm_py_bootstrap, sizeof rm_py_bootstrap - 1) < 0) {
```
fermino, line 179: BTW, I think `sizeof(rm_py_bootstrap) - 1` would be better for readability, but in this case it's a matter of taste.

vassilit:
> Looks good to me! Do you want this merged asap or should we wait for something else? (I haven't fully understood what the idea about master/develop is so I don't want to mess anything up :p)

It's not important, this commit is a detail. However, I think more effort should be done on the GUI installation (be convenient for package maintainers, make sure that the Python package is well installed, etc) before release :(

fermino (approved):

lib/cmdline.c
vassilit: I don't mind changing it, there are inconsistency in the rest of the code, but most often than not it is used with parenthesis.
I changed RM_PY_BOOTSTRAP to lower-case because usually uppercase is for preprocessor macros. But again, that's a matter of taste.

fermino: Yeah, I agree on the casing.

fermino (approved):

PR #712, opened 2025-03-11
simplify LFS

vassilit:


vassilit:
This will probably fail on 32bit systems. Many checks to do up until 2.11.0.
AIX and Irix 32-bit could fail as well.

vassilit:
> > This will probably fail on 32bit systems. Many checks to do up until 2.11.0. AIX and Irix 32-bit could fail as well.
> 
> I am fine with the updates but I have no way of checking this. Do we have issues for it so it doesn't got forgotten?

You can test 32-bit with `qemu-system-i386`. It works well, but it breaks compatibility with older Glib, as 2.64 is required under the new code.

vassilit:
Our FTS contrib lib does not support LFS properly.

[libc commit](https://sourceware.org/git/?p=glibc.git;a=commit;h=8b7b7f75d91f7bac323dd6a370aeb3e9c5c4a7d5)

RayOei:
This one needs rebasing on `develop`?

vassilit:
@RayOei this one is ready as well

RayOei (approved):
👍

PR #716, opened 2025-03-16
fix 32-bit compilation on old Glib

vassilit:
The 64-bit released version of rmlint **already** uses [G_SIZEOF_MEMBER](https://docs.gtk.org/glib/func.SIZEOF_MEMBER.html), that is available on 2.64. So in practice, rmlint 2.10.2 has already 2.64 as a minimum version.
However, the 32-bit code does not use this macro, so it still compiles under older Glib on 32-bit. Let's not break user compatibility for 2.10.

This will be properly fixed in 2.11.


vassilit:
> > This will be properly fixed in 2.11.
> In which issue is this included?

See https://github.com/sahib/rmlint/pull/708 and https://github.com/sahib/rmlint/pull/707

This one should be merged for 2.10.3.

RayOei (approved):

PR #717, opened 2025-03-17
Update badges

RayOei:
Badges tweaked

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,4 +1,4 @@
-name: Build and test rmlint (master)
+name: Build & test
 
 on:
   pull_request:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,4 +1,4 @@
-name: Build & test
+name: Build and test
 
 on:
   pull_request:
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 .. image:: https://raw.githubusercontent.com/sahib/rmlint/develop/docs/_static/logo.png
-  :align: center
+   :align: center
 
 `rmlint` finds space waste and other broken things on your filesystem and
 offers to remove it.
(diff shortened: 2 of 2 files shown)


vassilit:
Is there any way not to have the last badge on a new line ?

README.rst
```diff
@@ -61,8 +61,8 @@ your favourite distribution. If not, you might consider
 
 
 .. image:: https://repology.org/badge/vertical-allrepos/rmlint.svg?columns=5
- :target: https://repology.org/project/rmlint/versions
-  :align: center
+   :align: center
+   :target: https://repology.org/project/rmlint/versions
```
vassilit, line 65: should we align double-columns ?

vassilit: the initial version was better, let's merge the initial version

RayOei:
> Is there any way not to have the last badge on a new line ?

Not directly, that I am aware of. I did tinker a bit with putting them in a table but that looked a bit overdone. Also didn't spend too much time on it, TBH.

README.rst
RayOei: What do you mean with the initial version? The long single column?
In this particular case I changed the order so the target wouldn't include the alignment in the URL.

RayOei:
Actually: if we would rename the Github workflow to `build & test` it should fit on one line.

README.rst
vassilit: initial:
```
 :target: https://repology.org/project/rmlint/versions
  :align: center
```

new:
```
   :align: center
   :target: https://repology.org/project/rmlint/versions
```

.github/workflows/build-and-test.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and test rmlint (master)
+name: Build & test
```
vassilit, line 1: "&" sign mean anchor in YAML.

README.rst
RayOei: That initial variance leads to a URL that reads: `https://repology.org/project/rmlint/versions:align:center` which is invalid and leads to a `Page not found` error when you click the badge.
But I am also wondering what is `better` about it because for me it looks the same? 
Apart from the invalid URL.

.github/workflows/build-and-test.yml
RayOei: Oops 😊

RayOei:
BTW, any objections when I add `*.rst` to the build skip test filter? I noticed all these corrections are executing the tests too.

.github/workflows/build-and-test.yml
RayOei: That looks better and is now on one line

vassilit (approved):

PR #719, opened 2025-03-19
gui: migrate to gtksourceview4

vassilit:
Closes: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1079423

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,7 +44,7 @@ Here's a list of readily prepared commands for known operating systems:
     # Optional dependencies for building documentation:
     $ sudo dnf install python3-sphinx
     # Optional dependencies for the GUI:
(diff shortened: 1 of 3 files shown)


vassilit:
I think there is, the questions are, how old are they and do we want to be included in the upcoming Trixie?

fermino:
Yeah, I don't like the idea of holding it back either. AFAIK gtksourceview4 could be also co-installed with gtksourceview3 so that's a yes for me. @RayOei any thoughts?

RayOei:
Fine to me. If we want to be ready for Trixie it is an obvious step. If nothing breaks, obviously 😁 Haven't checked, yet. Will quickly give it a spin later today.

RayOei (approved):
Work fine for me on Ubuntu 24.04

PR #731, opened 2025-03-25
CI: run tests with musl in addition to glibc

fermino:
I've been playing around with adding support for musl. Although using docker is not necessarily the way to go, I think it's not "that" bad. Of couse other OSs will be another thing.


vassilit:
There is now a `/Containerfile` that provides some sort of testing on musl.

You could use this to test in CI ?

vassilit:
Completed in #807

PR #748, opened 2025-07-06
Add ZFS filesystem reflink support

### LINKED ISSUE #642: Only original_cmd in shell script generated for reflink
I'm trying rmlint 2.9 and latest git master (d9c80478a4bb1cd3ecfeecd1021ee28995c07134) on a ZFS filesystem with reflink support and am seeing the same issue as reported by others in the new closed #407 issue.

The binary is linked with blkid but the output still only contains `original_cmd`.

It works fine when using `sh:handler=hardlink` and when I edit the generated file using `perl -pe 's/cp_hardlink(?!\()/cp_reflink/g' rmlint.sh > rmlint_new.sh` it reflinks successfully.

### LINKED ISSUE #745: sh:handler=reflink does not work on zfs, even though zfs now supports cp --reflink
Title says all.

It is possible to make a file share extents using cp --reflink. The filesystem does not yet support the FIDEDUPERANGE kernel call, however sh:handler=reflink does not use this feature, so handler=reflink would work on zfs.

I suspect https://github.com/sahib/rmlint/blob/master/lib/utilities.c#L582 is relevant here.

Mic92:
ZFS now supports file cloning/reflinks via cp --reflink in OpenZFS 2.2+. This commit adds proper support for ZFS in rmlint by distinguishing between filesystems that support the FIDEDUPERANGE ioctl (used by 'rmlint --dedupe' and sh:handler=clone) and those that only support cp --reflink.

Changes:
- Split filesystem detection into fs_supports_dedupe() and fs_supports_reflinks()
- Add separate dedupefs_table to track filesystems supporting FIDEDUPERANGE
- Update clone handler to check dedupe support specifically
- Add rm_mounts_can_dedupe() function for checking dedupe capability
- Document that ZFS supports reflinks but not the dedupe ioctl

This ensures that:
- sh:handler=reflink works correctly on ZFS (uses cp --reflink)
- sh:handler=clone is skipped on ZFS (no FIDEDUPERANGE support)
- The 'link' shortcut now works properly on ZFS by falling back to reflink

Documentation has been updated to clarify the behavior on ZFS.

Fixes: #745

### DIFF
--- a/lib/utilities.c
+++ b/lib/utilities.c
@@ -579,7 +579,8 @@ static RmMountEntry *rm_mount_list_next(RmMountEntries *self) {
     }
 }
 
-static bool fs_supports_reflinks(char *fstype, char *mountpoint) {
+static bool fs_supports_dedupe(char *fstype, char *mountpoint) {
+    /* Check if filesystem supports FIDEDUPERANGE ioctl for deduplication */
     if(strcmp(fstype, "btrfs")==0) {
         return true;
     }
@@ -592,6 +593,49 @@ static bool fs_supports_reflinks(char *fstype, char *mountpoint) {
         int res = system(cmd);
         g_free(cmd);
         return(res==0);
+    }
+    return false;
+}
+
+static bool fs_supports_reflinks(char *fstype, char *mountpoint, char *fsname) {
+    /* Check if filesystem supports any form of reflinks/clones */
+    if(fs_supports_dedupe(fstype, mountpoint)) {
+        /* If it supports dedupe ioctl, it supports reflinks */
+        return true;
+    }
+    if(strcmp(fstype, "zfs")==0) {
+        /* ZFS supports reflinks/clones in newer versions (OpenZFS 2.2+).
+         * Note: ZFS does not support the FIDEDUPERANGE ioctl used by
+         * 'rmlint --dedupe', but does support 'cp --reflink=always'.
+         * For sh:handler=reflink, the generated script will use cp --reflink.
+         * For sh:handler=clone, the dedupe operation will fail gracefully. */
+
+        /* Check that the zpool has block_cloning feature enabled.
+         * Get the pool name from fsname (the part before the first '/', if any).
+         * It can be disabled, or not supported because the pool has not been
+         * upgraded yet, or because ZFS is too old. */
+        char *fs = g_strdup(fsname);
+        g_strdelimit(fs, "/", '\0'); /* modifies 'fs' inplace */
+
+        gchar *argv[] = {"zpool", "get", "-H", "-o", "value", "feature@block_cloning", fs, NULL};
+        gchar *standard_output = NULL;
+        gint exit_status = 0;
+        bool supports_cloning = false;
+
+        if(g_spawn_sync(NULL, argv, NULL, G_SPAWN_SEARCH_PATH | G_SPAWN_STDERR_TO_DEV_NULL,
+                        NULL, NULL, &standard_output, NULL, &exit_status, NULL)) {
+            if(exit_status == 0 && standard_output != NULL) {
+                /* Check if the feature is "active" or "enabled" */
+                if(strstr(standard_output, "active") != NULL ||
+                   strstr(standard_output, "enabled") != NULL) {
+                    supports_cloning = true;
+                }
+            }
+            g_free(standard_output);
+        }
+
+        g_free(fs);
+        return supports_cloning;
     }
     return false;
 }
@@ -675,13 +719,21 @@ static RmMountEntries *rm_mount_list_open(RmMountTable *table) {
                   evilfs_found->name, wrap_entry->dir, (unsigned)dir_stat.st_dev);
         }
 
-        if(fs_supports_reflinks(wrap_entry->type, wrap_entry->dir)) {
+        if(fs_supports_reflinks(wrap_entry->type, wrap_entry->dir, wrap_entry->fsname)) {
             RmStat dir_stat;
             if(rm_sys_stat(wrap_entry->dir, &dir_stat) == 0) {
                 g_hash_table_insert(table->reflinkfs_table,
                                     GUINT_TO_POINTER(dir_stat.st_dev),
                                     wrap_entry->type);
                 rm_log_debug_line("Filesystem %s: reflink capable", wrap_entry->dir);
+                
+                /* Also check if it supports dedupe ioctl */
+                if(fs_supports_dedupe(wrap_entry->type, wrap_entry->dir)) {
+                    g_hash_table_insert(table->dedupefs_table,
+                                        GUINT_TO_POINTER(dir_stat.st_dev),
+                                        wrap_entry->type);
+                    rm_log_debug_line("Filesystem %s: dedupe capable", wrap_entry->dir);
+                }
                 continue;
             }
         }
@@ -712,6 +764,7 @@ static bool rm_mounts_create_tables(RmMountTable *self, bool force_fiemap) {
     /* Mapping dev_t => true (used as set) */
     self->evilfs_table = g_hash_table_new(NULL, NULL);
     self->reflinkfs_table = g_hash_table_new(NULL, NULL);
+    self->dedupefs_table = g_hash_table_new(NULL, NULL);
 
     RmMountEntry *entry = NULL;
     RmMountEntries *mnt_entries = rm_mount_list_open(self);
@@ -838,6 +891,7 @@ void rm_mounts_table_destroy(RmMountTable *self) {
     g_hash_table_unref(self->nfs_table);
     g_hash_table_unref(self->evilfs_table);
     g_hash_table_unref(self->reflinkfs_table);
+    g_hash_table_unref(self->dedupefs_table);
     g_slice_free(RmMountTable, self);
 }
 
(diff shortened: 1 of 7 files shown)


jdehaan:
I'm not very familiar with the codebase here, but formally there are tests and maybe something needs to be adapted here as well to provide/enable testing?

https://github.com/sahib/rmlint/blob/a7e1df4b22a1be67597d46438193132464cf27d2/tests/utils.py#L49

speed47:
Damn, I just re-implemented it on my side before seeing that this PR was open... I should really have checked first!

I didn't go as far as differentiating between FIDEDUPERANGE and FICLONERANGE, but clearly that's a good thing to do.
However I implemented a check, as done with xfs, to ensure that block cloning is actually enabled/available on the given zpool. This is the code, taken from my would-have-been PR that you might want to add on top of yours if you want to add this check too (my local compiled version is actually your branch with this snippet on top):

```diff
diff --git a/lib/utilities.c b/lib/utilities.c
index ef52d347..a3c50c94 100644
--- a/lib/utilities.c
+++ b/lib/utilities.c
@@ -597,7 +597,7 @@ static bool fs_supports_dedupe(char *fstype, char *mountpoint) {
     return false;
 }

-static bool fs_supports_reflinks(char *fstype, char *mountpoint) {
+static bool fs_supports_reflinks(char *fstype, char *mountpoint, char *fsname) {
     /* Check if filesystem supports any form of reflinks/clones */
     if(fs_supports_dedupe(fstype, mountpoint)) {
         /* If it supports dedupe ioctl, it supports reflinks */
@@ -609,7 +609,20 @@ static bool fs_supports_reflinks(char *fstype, char *mountpoint) {
          * 'rmlint --dedupe', but does support 'cp --reflink=always'.
          * For sh:handler=reflink, the generated script will use cp --reflink.
          * For sh:handler=clone, the dedupe operation will fail gracefully. */
-        return true;
+
+        /* However we need to check that the zpool has the feature enabled,
+         * so get the pool name from the fsname (this is the part before the
+         * first '/', if any). It can either be disabled, or not supported
+         * because the pool has not been upgraded yet, or because ZFS is too old. */
+        char *fs = g_strdup(fsname);
+        g_strdelimit(fs, "/", '\0'); /* modifies 'fs' inplace */
+        char *pool = g_shell_quote(fs);
+        char *cmd = g_strdup_printf("zpool get -H -o value feature@block_cloning %s 2>/dev/null | grep -q -e active -e enabled", pool);
+        int res = system(cmd);
+        g_free(cmd);
+        g_free(pool);
+        g_free(fs);
+        return(res==0);
     }
     return false;
 }
@@ -693,7 +706,7 @@ static RmMountEntries *rm_mount_list_open(RmMountTable *table) {
                   evilfs_found->name, wrap_entry->dir, (unsigned)dir_stat.st_dev);
         }

-        if(fs_supports_reflinks(wrap_entry->type, wrap_entry->dir)) {
+        if(fs_supports_reflinks(wrap_entry->type, wrap_entry->dir, wrap_entry->fsname)) {
             RmStat dir_stat;
             if(rm_sys_stat(wrap_entry->dir, &dir_stat) == 0) {
                 g_hash_table_insert(table->reflinkfs_table,
```

I'm not extremely fond of using the shell version of `system()`, but I didn't want to introduce a different way of doing it w.r.t the xfs check.

CodingWithAnxiety:
Hey, Finally back from my hiatus.

I'll give this a readover, though as @jdehaan says, unit tests should probably be added for this. I may commit to this in the future with this.

EchterAgo:
You might want to use `--reflink=auto` instead of `--reflink=always` to make this work across different datasets. See https://github.com/openzfs/zfs/issues/15345

lib/utilities.c
```diff
@@ -596,6 +597,49 @@ static bool fs_supports_reflinks(char *fstype, char *mountpoint) {
...
+         * upgraded yet, or because ZFS is too old. */
+        char *fs = g_strdup(fsname);
+        g_strdelimit(fs, "/", '\0'); /* modifies 'fs' inplace */
+
+        gchar *argv[] = {"zpool", "get", "-H", "-o", "value", "feature@block_cloning", fs, NULL};
+        gchar *standard_output = NULL;
+        gint exit_status = 0;
+        bool supports_cloning = false;
+
+        if(g_spawn_sync(NULL, argv, NULL, G_SPAWN_SEARCH_PATH | G_SPAWN_STDERR_TO_DEV_NULL,
```
vassilit, line 625: An alternative would be not to check reflink-capability at all here and let the _coreutils_ commands used in the script to emit error if the ioctl calls fail.

cp seems to [try](https://github.com/coreutils/coreutils/blob/95044cb5eaea83d02f768feb5ab79fcf5e6ad782/src/copy.c#L1008) for `FICLONE` by default and use copy_file_range on failure.

Our `fs_supports_dedupe()` uses `system()` for XFS so what we do here is not so bad, at least we don't invoke the shell.

gmelikov:
@vassilit saw your commit with zfs tests, iirc ubuntu 24 has zfs 2.2, it has disabled block cloning by default. It may be enabled with `echo 1 > /sys/module/zfs/parameters/zfs_bclone_enabled`. Some tests _may_ not work because in zfs 2.2.2 block cloning may be not 100% working, in this case feel free to ping me, I can add fresh zfs build to CI if you want it (basically it's described here https://openzfs.github.io/openzfs-docs/Developer%20Resources/Custom%20Packages.html#debian-and-ubuntu ).

vassilit:
@gmelikov Спасибо! But I saw your message too late (just now).
I am running ZFS extensively on FreeBSD so I naively tried sysctl :) then modprobe zfs zfs_bclone_enabled=1 (strangely enough, seems to be the default in `man 4 zfs` on ZFS 2.2.2 on Ubuntu 24.04.3 (the one used on runners), but is not de facto enabled) and then settled for using the sys filesystem as postinst of _zfsutils-linux_ do `insmod zfs` anyway.

I plan to add FreeBSD tests on 14.4 (ZFS 2.2) and 15.0 (2.4) soon that will cover more recent versions if 2.2.2 on Linux is failing.

gmelikov:
@vassilit usually get FreeBSD in github actions is not so straightforward, but yeah, it's too should work :) Although better to have both Linux and FreeBSD. JFYI openzfs in the end just started qemu in CI for non-Ubuntu OSes https://github.com/openzfs/zfs/blob/master/.github/workflows/zfs-qemu.yml

vassilit:
@gmelikov it's a bit off-topic here but for FreeBSD runners on Github I used Cirrus-CI in the past, then switched to [vmactions](https://github.com/vmactions/freebsd-vm), but it is very slow. The quite new [Firecracker](https://github.com/acj/freebsd-firecracker-action) seems promising but I have not tested it yet.

vassilit:
> Some tests _may_ not work [...], in this case feel free to ping me, I can add fresh zfs build to CI if you want it 

@gmelikov, ping. Your help is appreciated, could you open a separate PR (I don't know how to give you rights on Mac92's branch) and we will rebase this one on yours ?

gmelikov:
@vassilit got it, I'll do that in several days.

gmelikov:
zfs 2.4 didn't help, and looks like problem is not with zfs 2.2 from Ubuntu, here you can see run with 2.4: 
- https://github.com/gmelikov/rmlint/actions/runs/20502566898/job/58912158085
- commit to build zfs 2.4 https://github.com/gmelikov/rmlint/commit/e6a44fd20b637e02cded22c8690580fa2fc8068b (it's WIP, because it didn't help)

I've built this PR, read `test_second_extent_differs` test and tried by hand on zfs 2.2 (`/tmp` is ZFS dataset):
```bash
ubuntu@ubuntu:~/rmlint$ sudo zfs -V
zfs-2.2.2-0ubuntu9.2
zfs-kmod-2.2.2-0ubuntu9.1
ubuntu@ubuntu:~/rmlint$ echo 1 | sudo tee /sys/module/zfs/parameters/zfs_bclone_enabled
1

# Create original file
ubuntu@ubuntu:~/rmlint$ dd if=/dev/urandom of=/tmp/a bs=4K count=1

# Before: block table empty
ubuntu@ubuntu:~/rmlint$ sudo zdb -T rpool
BRT: empty

# Run test
ubuntu@ubuntu:~/rmlint$ cp --reflink=auto /tmp/a /tmp/b

# After, ZFS sees duplicates, reflink was a success
ubuntu@ubuntu:~/rmlint$ sudo zdb -T rpool
BRT: used 4K; saved 4K; ratio 2.00x

ubuntu@ubuntu:~/rmlint$ ./rmlint --is-reflink /tmp/a /tmp/b
ubuntu@ubuntu:~/rmlint$ echo $?
5
```

from docs exit code 5 is "fiemaps can't be read". Ah, I thought ZFS supported it but it's not https://github.com/openzfs/zfs/pull/9554

So, reflink works  apparently, BUT rmlint can't  easily test it with FIOMAP.

vassilit:
2.4 helped not getting EAGAIN: Resource temporarily unavailable. At least, we can test now.

 Thank you !

gmelikov:
@vassilit ok, here is clean commit to use zfs 2.4 https://github.com/gmelikov/rmlint/commit/86ed25491c8055688e6f7401557b9e952e152b78

vassilit:
I guess we could add _needs_fiemap_ to some tests and modify the code so that reflink operations are best-efforts on ZFS (no easy way to check, I don't think we should call zdb or import headers from OpenZFS and use their non-stable API).

It is kind of already supported, as we can make the generated script to use `--reflink=always` on Linux and on FreeBSD it's not clear for me how to call  `copy_file_range(COPY_FILE_RANGE_CLONE)`, `clone_file_range`, `dedupe_file_range` and `remap_file_range` from 15.0-RELEASE base userland utilities. That may change with future versions.
No idea on other platforms that uses ZFS and if they ever support block cloning (Illumos-based, Solaris-based, etc)

What are your thoughts on this?

chrisportela:
Recently ran this on version a ZFS v2.3.4 volume with block_cloning enabled and it told me i would have ~1TB of space savings, but it doesn't appear to have freed up as much space as I expected. I only have 169GB free after running the rmlint.sh script.

Am I misunderstanding something about ZFS and the reflink support added here?

intelfx:
@chrisportela due to insufficiently flexible Linux kernel interfaces, ZFS only supports reflinks via FICLONE ioctl within one dataset. Cloning between different datasets (_provided_ those are actually physically compatible, i.e. same recordsize) is only possible via `copy_file_range()` ioctl. The problem is that `cp --reflink=always` (which is what rmlint emits) only uses FICLONE, because FICLONE guarantees a reflink or fail by contract. The `copy_file_range()` syscall (which is a best-effort cloning operation by its contract) is only used in `cp --reflink=auto`.

I have a huge pile of fixes to the entire reflinking feature in my pipeline; perhaps I'll also add a slightly-smarter zfs support before I send it out.

vassilit:
> The problem is that `cp --reflink=always` (which is what rmlint emits) only uses FICLONE, because FICLONE guarantees a reflink or fail by contract. The `copy_file_range()` syscall (which is a best-effort cloning operation by its contract) is only used in `cp --reflink=auto`.

Please note that on FreeBSD, `copy_file_range()` with the COPY_FILE_RANGE_CLONE flag also garantees a reflink of fail by contract.

PR #755, opened 2025-09-23
fix spurious NUL char in cksum xattr

### LINKED ISSUE #744: Is the \0 stored in the xattrs for a checksum deliberate?
I was doing some programming that involved storing md5 checksums in a file's xattr, and I noticed that rmlint's storage of a checksum includes a 0 byte, which causes the listing (by xattr -lz and other commands) to be displayed in hex.  I presume this happens because rmlint issues a setxattr call from C to save user.rmlint.md5.cksum in an xattr, and passes a length that includes the null byte of the string, whereas when I set user.checksum.md5 from the command line, the setfattr command-line command does not include the NUL byte?

user.rmlint.md5.cksum:
0000   30 39 65 33 30 61 31 35 65 35 61 37 34 64 65 61    09e30a15e5a74dea
0010   62 64 33 34 66 30 34 66 35 64 30 37 33 38 37 36    bd34f04f5d073876
0020   00                                                 .

user.checksum.md5: 09e30a15e5a74deabd34f04f5d073876

I'm just wondering if this was a deliberate choice in the rmlint code and if so, why?  If it was unintentional, would removing the terminating NUL byte in an updated release break backwards compatibility?  It would certainly make the xattrs more readable and more easily used in other scripts.

thanks,

Graham

speed47:
`sizeof(string)` used instead of `strlen(string)` added a spurious NUL character in the xattr, remove it. closes #744.

### DIFF
--- a/lib/xattr.c
+++ b/lib/xattr.c
@@ -234,7 +234,7 @@ int rm_xattr_write_hash(RmFile *file, RmSession *session) {
     if(rm_xattr_build_key(session, "cksum", cksum_key, sizeof(cksum_key)) ||
        rm_xattr_build_key(session, "mtime", mtime_key, sizeof(mtime_key)) ||
        rm_xattr_build_cksum(file, cksum_hex_str, sizeof(cksum_hex_str)) <= 0 ||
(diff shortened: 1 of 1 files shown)


vassilit (approved):
Indeed :)

thx

PR #773, opened 2026-07-15
Merge master into develop

### LINKED ISSUE #765: Is she Dead Jim?
Once again just checking if this project is dead?
There's been no commits for more than half a year if I see the stats correctly.  last tag was first quarter of last year.

Is she dead?

vassilit:
Two thirds manually done about a year ago and abandonned.
One third recently resumed, on and off, with some targeted AI (various models) assistance.

All pedantic tests pass (~6 hours on tmpfs+BTRFS ramdisk with a 4GHz AVX512-enabled 32-core CPU).

After this PR is rebased (I hope Github will let us do a clean one *) into develop, the plan is to rebase any PRs targeting develop (I've just created a new label for them), test each one, and merge them as appropriate.

After additional testing of develop, we'll merge it into master and release 2.11.0.

*: fellow maintainers, after approving, please don't merge, but rebase. And don't rebase from Github WUI, but from the CLI, verify it looks good and push. If you're unsure, I'd be happy to handle the rebase myself.


RayOei:
Hi @vassilit, 
I don't have time this weekend to look at it so I only did a quick scan:
* the workflow is complaining a bit: it seems `\actions/checkout@v4` should now be `\actions/checkout@v7`  
* isn't `dev` already ontop of `develop`? The title suggests you want to merge `master`?

How much reviewing, or which parts, would you want to focus on?

vassilit:
There used to be commits not directly related to the merge in this PR. I put them in separate PRs for independant reviews and should be merged later on:

- modernising of the test suite: #774 
- small changes on CI: #775 
- GUI: cleaning and compat with modern lib: #776 (+ one hotfix for master on a separate branch for modern distributions)

This PR #773 is essentially four commits:
- one cherry-pick commit of a compilation fix from master. Included before the merge to make sure that we can always compile a version of develop just before the merge, to bisect/cross-check eventual inconsistencies in the future. After the merge, this is essentially a no-op (as the fix is already on master).
- a merge commit integrating master's changes and fixes to current develop, almost bug-for-bug.
- one commit fixing a bug (always present in master, but side-effects made it invisible) that emerged during the tests, due to develop changing an enum's order. A more precise test is added
- one commit fixing a testing bug, that was present for a long time during pendantic runs

So the idea was to bring the PR to a fully testable state (including pendantic).
That's it. Everything else is now in others PR, that targets dev for now, but will target develop once rebased.

> * the workflow is complaining a bit: it seems `\actions/checkout@v4` should now be `\actions/checkout@v7`

master and develop both agrees on v4, so this stays on v4 here but is upgraded to v7 in #775.

> * isn't `dev` already ontop of `develop`? The title suggests you want to merge `master`?

`dev` is based on top of `develop`, so `develop` is older. This PR is to make `develop`'s HEAD the same as `dev`.
`dev` is `develop` + `master` so merging this PR in develop will make `develop` = old `develop` + `master` on top of it.

In the future, merging `develop` to `master` should be conflict-less.

> How much reviewing, or which parts, would you want to focus on?

I’ve already done a lot of testing (including pendantic runs with and without AVX512 flags), but extra eyes are always welcome.
However, as 'develop' is a development branch (not released per se), it will always be possible to fix things later on, so I wouldn't lose too much time on it now. Once merged, devs and "beta-testers" (Arch has a develop-git package) will help I'm sure.
One thing to do would be to run the testsuite on your platform, some things differs between systems. If it fails on this branch but succeed on develop and master, then we should correct this PR. If it fails on both master or develop and on this PR, this is normal.
Tests that are not covered right now but that we plan(?) to add in the future (i.e. 32 bit) are to be dealt with later.

We cannot merge complex PRs into `master` for fear of increasing the divergence between branches. Once this PR is merged, the previous organisation will be restored, and new development will need to take place on `develop`.
Nevertheless, I think the idea of ​​keeping this product stable remains IMHO. New devs should involve first fixing bugs, test coverage, quality, minor-to-medium additions (extending support to ZFS), external library upgrades (and their interface) to ensure that rmlint still compiles on Linux distributions in 2030.

RayOei:
Ofcourse the M.2 SSD died in my test rig...

So I have now used a Multipass Ubuntu VM on my Mac. I am running into some tests which are failing and I yet have to look into those. I also ran into an issue with `scons` in which the pgk-config call fails when no version is defined.  E.g. `conf.check_pkg('json-glib-1.0', 'HAVE_JSON_GLIB', required=True)` fails while `conf.check_pkg('json-glib-1.0 >= 1.10', 'HAVE_JSON_GLIB', required=True)` succeeds. Anything that rings a bell?

I am obviously not fully sure this is not related to my specific setup. I intend to rebuilt my Linux machine as soon as the new SSD arrives.

vassilit:
Hello @RayOei,

I'm sorry for your SSD. I hope that rmlint tests are not the culprit. They can be pretty disk intensive, that's why it's recommended to run them in volatile memory.

The `check_pkg` failure does not ring a bell. 
If you have json-glib 1.10.8, `pkg-config --exists 'json-glib-1.0'; echo $?` and `pkg-config --exists 'json-glib-1.0 >= 1.10'; echo $?` should have the same result.
I've already began a bit of rework of SCons scripts, I'll look into it.
That's probably due to your setup, as the configure stage is CI-tested on four versions of Ubuntu and they pass.

The question for this PR is: do `scons config` works on develop (before the merge) but does not on dev (after the merge) ?
If they both don't work, I would consider the tests successful, as we are only testing regressions here.

RayOei:
No worries... my SSD died before I could start working.
As for the pkg-config: I have added both to sconstruct (I hadn't done that) and they return different results. Only the one with the version check succeeds. Both checks _do_ succeed in the terminal so it seems a `TryAction` issue?
I will continue with the merge (locally obviously) to compare the tests. There are only some failing, for instance an assert expects a string to end with `x00` while the result has the correct string but without the null byte. I will not dive into the test, yet ;-)

vassilit:
Which version of scons do you use ?

The failing test in test_cache.py is a real (test) regression. It is about #755, which forgot to update the tests.
I'll update the ci to run the tests as root.

RayOei:
I have tried scons 4.8.1 and the latest 4.10.1. Both have the same result.
Ah...#755 does explain that, indeed. There are some others failing too, though (not many). But I will look into that later. First comparing the before and after merge.

RayOei:
BTW. I also checked the other `pkg-config` calls without a version and **not** required: they fail as well (checking the rc) but that is not reported, when set to required it is shown that those are not found. Adding a (minimum) version solves it, obvously.

vassilit:
do you have `$PKG_CONFIG` set to something in your environ ? 
My guess is SCons is not executing `pkg-config`, but something else.

Do it shows exactly `Error: json-glib-1.0 not found.` ? Or is SCons is not displaying the print, `Checking for json-glib-1.0...` and then it exit with an error ?

Regarding the failed tests, I've run the test suite as root from master, develop and dev. All xattr fails on tests/test_options/test_cache.py::test_xattr_detail, develop on a different assertion.

But only develop does not fail on `tests/test_robustness/test_path_doubles.py::test_mount_binds`.
I'll look into it.

For the record:
- dev: `3 failed, 176 passed, 13 skipped in 1315.88s (0:21:55)`
- master: `3 failed, 171 passed, 12 skipped in 1209.65s (0:20:09)`
- develop: `2 failed, 152 passed, 8 skipped in 1140.44s (0:19:00)`

Tests that differs:
All 186 of master's tests are present on dev.
All develop's tests are present on dev, some of which were parametrized: test_stdin.py::test_stdin_empty and test_merge_directories.py::test_simple.
Added test_others.py::test_traversed_dirs_are_no_lint in a separate commit.

About --is-reflink comment, my comment stating that we got it wrong ... got it worng. On master `--is-reflink` uses rm_cmd_parse_args() calls rm_cmd_set_paths(), which calls rm_cfg_add_path(), which calls realpath(). So by the time rm_util_link_type() lstat's the path, the path has already been rewritten. So on master we resolved symlinks indeed.

RayOei:
Nope.. no environment set.
I work from a new Ubuntu 26.04 btw.
As far as I can see pkg-config is a symlink to pkgconf v2.5.1. I assume that is what Ubuntu installs by default.
There are no other references to pkg-config. 
The issue is solved by adding the version check, so it works that way. But it is puzzling...

I am running with the long, long Valgrind/pendatic settings just to see how that pans out. Still running :-)

vassilit:
you are not on an AVX512 platform I guess ? Because valgrind have some problem with it.

I will rework the merge a bit on the xattr and mount_binds issues. 
Seems like xattr is failing on mtime tolerance.

RayOei:
Valgrind takes forever ;-)

Fast settings:
dev - 1 failed, 177 passed, 14 skipped in 706.98s (0:11:46) 
develop merge dev -  1 failed, 177 passed, 14 skipped in 733.71s (0:12:13)

Seem to be getting slightly different results, maybe related to diff env? I have no CPU extensions for one.
Either way: for this stage it seems fine enough?

RayOei (approved):
;-)

vassilit:
Thanks for the tests.

There were no tests that succeeded in both master and develop and failed here.

PR #774, opened 2026-07-16
tests: modernisation

vassilit:
We now use tools that the system find in PATH.

(if dev is merged to develop before merging this PR, modify the destination branch here)


docs/testing.rst
RayOei, line 1: I would add the requirement of pytest >= 9.
By default my system installed 8.4.2 apparently. So I encountered issues with the marker. Took me a sec (or two) to realise it was the version.
With `pip install -U pytest` I got v.9.1.1 and the `slow` marker worked

RayOei (commented):
With `-n auto` 171 passed, 9 skipped in 18.98s
Without `-n auto` 171 passed, 9 skipped, 7 deselected in 84.36s (0:01:24) 

Noticed several fixes 👍 of things I also collected the past days. 
Looks great. Specific things you want to check?

vassilit:
`RM_TS_PEDANTIC=1 pytest tests/test_formatters/test_others.py::test_just_call_it` and 
`RM_TS_PEDANTIC=1 pytest tests/test_formatters/test_py.py::test_paranoia`
were failing. This is fixed in the last commit.

vassilit:
I think this is ready to be merged if the tests are successful on your side.
(you can merge (not rebase) any time, no merge commit in this batch)

RayOei (approved):
176 passed, 10 skipped, 1 error 
With pedantic. The only one failing is `test_collision_resistance`
I assume that is fine for now. I will dig a bit deeper to check what is the case there, it seems the underlying `make_numbered_dir` fails.

Issue #771, opened 2026-06-19
Claude Code Review Report


qarmin:
Recently, I ran Claude Code across several projects(like image-rs, lofty of symphonia) to look for potential issues(logic bugs, copy paste errors, architecture problems, panics etc.).

In this projects, a significant portion of reported findings turned out to be incorrect or minor false positives, but a non-trivial subset was valid and included real issues ranging from incorrect comments to actual logic bugs. As a result, the findings generally require manual validation to separate noise from actionable problems.

Full Report - [rmlint_20260619.html](https://github.com/user-attachments/files/29145486/rmlint_20260619.html)

Example findings (this is only a subset of the findings, specifically those most likely to be actual bugs; for the full list, see the report above):

=====================================

### CPY_1 `MEDIUM` ⚡ easy

**Description:** In rm_file_cmp_without_extension(), when computing the comparison length for file B's basename, the fallback (no-extension) branch incorrectly uses `strlen(basename_a)` instead of `strlen(basename_b)`: `size_t b_len = (ext_b) ? (ext_b - basename_b) : (int)strlen(basename_a);`. When file B has no extension and its basename length differs from file A's, b_len is wrong, which can cause incorrect SIGN_DIFF results and/or g_ascii_strncasecmp comparing the wrong number of characters under --match-without-extension. This can misclassify or fail to classify files as duplicates/originals based on basename matching.

**Locations:**

```
lib/preprocess.c:49-63
   49 | static gint rm_file_cmp_without_extension(const RmFile *file_a, const RmFile *file_b) {
   50 |     const char *basename_a = file_a->folder->basename;
   51 |     const char *basename_b = file_b->folder->basename;
   52 | 
   53 |     char *ext_a = rm_util_path_extension(basename_a);
   54 |     char *ext_b = rm_util_path_extension(basename_b);
   55 | 
   56 |     /* Check length till extension, or full length if none present */
   57 |     size_t a_len = (ext_a) ? (ext_a - basename_a) : (int)strlen(basename_a);
   58 |     size_t b_len = (ext_b) ? (ext_b - basename_b) : (int)strlen(basename_a);
   59 | 
   60 |     RETURN_IF_NONZERO(SIGN_DIFF(a_len, b_len));
   61 | 
   62 |     return g_ascii_strncasecmp(basename_a, basename_b, a_len);
   63 | }
```

[...]

PR #779, opened 2026-07-21
Various fixes

vassilit:


RayOei (approved):
1 failed, 177 passed, 10 skipped, 1 error 

As a general comment for running the test, at least on my system: 
It seems that between tests or groups some environment dependency exists.
I have been trying to run only subsets of tests and then `test_manylongpathfiles` (for instance) succeeds. Often some `tests/test_robustness/` tests fail. But run by itself most pass. 
The one that still fails, passes when run alone.

RayOei:
@vassilit I overlooked this test `def test_extensionless():`
The `assert footer['total_lint_size'] == 0` fails as the rmlint result is `3`.
It doesn't execute the next `assert footer['duplicates'] == 0` but shouldn't that be `1`? Or am I missing something here?

>[{"description":"rmlint json-dump of lint files","cwd":"/home/ubuntu/ramdisk/pytest/popen-gw2/test_extensionless0/","args":"/home/ubuntu/rmlint-orig/rmlint -V /home/ubuntu/ramdisk/pytest/popen-gw2/test_extensionless0 -i -o json:/home/ubuntu/ramdisk/pytest/popen-gw2/test_extensionless0/out.json -c json:oneline","version":"2.10.3","rev":"6bf93d2a","progress":0,"checksum_type":"blake2b","merge_directories":false},
> {"id":3923198470,"type":"duplicate_file","progress":100,"checksum":"b6f289a27d4fe90da63c503bfe0a9b761a8f76bb86148565065f040be6d1c3044cf7ded78ef800509bccb4b648e507d88dc6383d67642aadccce443f1534330a","path":"/home/ubuntu/ramdisk/pytest/popen-gw2/test_extensionless0/abc","size":3,"depth":1,"inode":15402054,"disk_id":55,"is_original":true,"twins":2,"mtime":1785074809.2655096},
> {"id":1555563892,"type":"duplicate_file","progress":100,"checksum":"b6f289a27d4fe90da63c503bfe0a9b761a8f76bb86148565065f040be6d1c3044cf7ded78ef800509bccb4b648e507d88dc6383d67642aadccce443f1534330a","path":"/home/ubuntu/ramdisk/pytest/popen-gw2/test_extensionless0/abcd","size":3,"depth":1,"inode":15402055,"disk_id":55,"is_original":false,"twins":2,"mtime":1785074809.2655222},
> {"aborted":false,"progress":100,"total_files":2,"ignored_files":0,"ignored_folders":0,"duplicates":1,"duplicate_sets":1,"total_lint_size":3}]

vassilit:
what's the rev number when doing `./rmlint --version` ?

RayOei:
@vassilit Forgot to actually _send_ the commend 😊 
Either way: I may have to revisit this as with my new machine I started fresh and don't seem to have any of these issues. As this was happening on my macOS/VM setup it may be a problem there.

vassilit:
> but shouldn't that be 1 ?

No, because we are using -i.

The manpage describe:
``-i --match-without-extension`` / ``-I --no-match-without-extension`` (**default**):

    Only consider those files as dupes that have the same basename minus the file
    extension. For example: ``banana.png`` and ``Banana.jpeg`` would be considered dupes,
    while ``apple.png`` and ``peach.png`` won't. The comparison is case-insensitive.

in the test, _abc_ and _abcd_ does not have the same **basename**, i.e. their name without extension. They don't have extension so their basename are actually the same as their full name and as they differs they should not be included in the duplicates list.

The bug/typo was we took the length of the first one always (basename_a), so before my patch, rmlint compared 'abc' with 'abcd' and _wrongly_ decided they have the same length, so they could be compared further only up to the length of 'abc'. And as they have the same prefix, they compare equal on their first three characters.

See CPY_1 in #771, it's explained in more details. 

And sorry not to have answered earlier, the rev number is actually in your output (6bf93d2a), which is before that PR, so somehow the tests have been updated on your working tree but not the binary or the code.

Now, I'm blocked by #799 (it is not big) which unfix the CI for all the others branches I'm working on (GUI, FreeBSD, macOS).
You don't need to test (but you absolutly can, that's helpful anyway), there is CI for that, but just have a general look at the diff.  The bigger manual testing would go before reaching master and tagging a release.
I would like to merge my unfinished work (take it out of my mind) so that I can rest a little.

PR #775, opened 2026-07-18
ci: modernise

vassilit:
- Compile with the CPU flags present on Github's VM
- upgrade checkout to v7

PR #780, opened 2026-07-24
SConstruct: fix completely wrong CPU feature detection

intelfx:
## tl;dr

Binary featureset must never depend on the build host's CPU.

## Motivation and explanation

Currently, scons build system inspects the build host's CPU (via `py-cpuinfo` which reads `/proc/cpuinfo`) and uses the result to
(a) add global `-m*` compiler flags, and
(b) exclude vendored blake3 SIMD source files and set `-DBLAKE3_NO_*` defines.

This is wrong for two complementary reasons:

- This overrides the environment's build flags, potentially activating ISA extensions beyond the intended baseline (`-mavx2` etc. leak into every TU, even if `$CFLAGS` is set to e.g. `-march-x86-64-v2`).
- This deactivates runtime-dispatchable code paths in blake3: even though blake3 contains its own cpuid-based runtime dispatch, `-DBLAKE3_NO_*` compiles those branches out and the matching object files are dropped, so a binary built on a host without AVX-512 can never use AVX-512 regardless of the CPU it actually runs on.

As a rule, the build system must never base any decisions on the build host's CPU.

Rework the build system such that every ISA-dependent implementation is either built unconditionally and selected via runtime dispatch, or gated on the effective target (the combination of `$CC` and `$CFLAGS`).

Additionally, make a proper distinction between C and assembly SIMD implementations. BLAKE3 has implementations in assembly for x86_64, and implementations in C with intrinsics for i386. Now, optimizing i386 builds might not be a priority, but BLAKE3's runtime dispatch expects symbols to exist when building for any x86, thus, build the assembly impls when building on x86_64, and C impls otherwise.

[Note for #775: when this is merged, py-cpuinfo will not be a dependency anymore]


vassilit:
You're right, it was particularly ridiculous because it seems that the relatively expansive CPUID was still called (via get_cpu_features) at runtime. If we could completly bypass runtime feature detection, I wouldn't call it «completly wrong» (unless I was a Fennoswedish programmer ;), there are cases where it's useful.

On i386, someone doing `CFLAGS="-march=native" scons config` might not get what expected, but anyway I think we should drop i386 support (while keeping 32bit ARM).

Also, I prefer tabs over spaces, really. But as all the existing codebase follows PEP recommandations and are using spaces, it's cleaner to keep it that way.

I had unfinished local branches that conflicts with your PR. I've pushed them in #783 and #784.
In them I've ignored Windows Blake3 support completly, as I don't think rmlint compiles or has been ported to Windows. If you've tested it, merci/спасибо, as I was very lazy to do it. But there are mac users and M2+ supports NEON.

Once the formatting will be 'fixed' (in quotes because I agree that indent with tab + alignement with space is nicer), I can either merge your branch and rebase mines or you rebase yours into mine. As you wish.

vassilit:
It would be a good idea to switch the default from blake2 to blake3 after this PR is merged.

intelfx:
My apologies for the delay—I saw that the new BLAKE3 checkout you've pulled in now supports NEON, realized that now I also need to test this on armv7l/aarch64, and got severely sidetracked (on the bright side, QEMU's UEFI firmware now supports virtio-input on ARM...)

> You're right, it was particularly ridiculous because it seems that the relatively expansive CPUID was still called (via get_cpu_features) at runtime. If we could completly bypass runtime feature detection, I wouldn't call it «completly wrong» (unless I was a Fennoswedish programmer ;), there are cases where it's useful.

The point was that we used compile-time CPU feature detection to inform flags that subsequently leaked into _all_ source files, not just BLAKE3. As a result, you **could not** produce a rmlint binary for older processors while building on a newer one.

This created actual practical problems: e.g. I have a Zen 3 machine in a colocation where I build all my packages (which is [x86-64 v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels:~:text=x86%2D64%2Dv3,-AVX)[^x86-64-v3]), and an old Intel Celeron N5105 NUC at my home (which is only [x86-64 v2](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels:~:text=cvtpi2pd-,x86%2D64%2Dv2)[^x86-64-v2]).

The problem is that it is *impossible* to build a rmlint package on the Zen 3 that would work on the Celeron (unless you just don't install py-cpuinfo). Even if I build rmlint with the correct `-march=x86-64-v2` (or with no `-march=` at all), the resulting binary would SIGILL on the Celeron. This is wrong however you look at it.

[^x86-64-v3]: x86-64-v3 is x86-64-v2 + AVX, AVX2, BMI1, BMI2, F16C, FMA, LZCNT, MOVBE, OSXSAVE
[^x86-64-v2]: x86-64-v2 is x86-64 baseline + CMPXCHG16B, LAHF-SAHF, POPCNT, SSE3, SSE4_1, SSE4_2, SSSE3

> Also, I prefer tabs over spaces, really. But as all the existing codebase follows PEP recommandations and are using spaces, it's cleaner to keep it that way.

Yep, sorry — my editor misdetected tabs vs spaces and I failed to double-check before committing. Of course indentation must be consistent.

> On i386, someone doing CFLAGS="-march=native" scons config might not get what expected, but anyway I think we should drop i386 support (while keeping 32bit ARM).

Why? I think they should get what is expected. With this PR, whatever `-march=` you specify will be used for all TUs as a baseline, and in addition to that we will forcibly add various `-m*` flags only when compiling C files with intrinsics. As for dropping i686, I don't have a preference, I just made everything work as it is now ;)

---

Anyway, I have rebased the branch just to prove that I'm alive, but I still need to make it work with ARM. I'll do it in a few hours.

vassilit (approved):

vassilit:
I will rebase #786 to your changes.

Neon and ARM were not CI-tested and we can add NEON support in a separate PR later on.

PR #783, opened 2026-07-25
build: modernise

vassilit:


RayOei (approved):
1 failed, 175 passed, 10 skipped, 1 error 
Fine for now. Any tips for the size of the ramdisk? Even with 20Gb it seems too small - according to the error.

vassilit:
Maybe there is a way to tune xdist not to run disk-space intensive tests at the same time.
Also, the some tests take a lot of inodes.

PR #784, opened 2026-07-25
modernise: Blake*

vassilit:

PR #785, opened 2026-07-25
Scons check

RayOei:
Add scons package versions
Fix path 

@vassilit If you don't mind adding these versions?

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -665,13 +665,13 @@ conf.env.Append(CCFLAGS=[
 ])
 
 conf.env['HAVE_GIO_UNIX'] = 0
-conf.check_pkg('gio-unix-2.0', 'HAVE_GIO_UNIX', required=False)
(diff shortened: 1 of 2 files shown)


vassilit:
Not really. 

But for the sake of it, could you try `rm -fr .sconsign.dblite .sconf_temp config.log` and run scons config again (without the versions) ?

vassilit:
Could you review #783 and #784 first ? As they will conflict.

RayOei:
Ha... the issue was with `.sconf_temp`. Some files in there didn't want to be deleted. Those were over a year old.
In that respect the version check is no longer _needed_ although it it may help others with a similar issue as me?

vassilit:
The problem is that it pins versions even if we don't need to. We don't use any interface that was not present when the lib went out.

535303d fixes `scons --config=force`, that regenerate the cache.

RayOei:
👍

PR #786, opened 2026-07-26
Miscelanous changes to SCons, drop SHA512-guard and linting

vassilit:


vassilit:
@RayOei

RayOei (approved):
= 176 passed, 6 skipped, 7 deselected =
You did _a lot_ of work here!
👍

vassilit:
Nice, because I've just wrote in the commit that pkg/ is planned for deletion to avoid conflicts and it became difficult trying to write commits in a way to avoid overlapping with others PR :)

vassilit:
Pushing again to allow Github to allow me a rebase onto develop (as they are somehow unrelated commit, they will go independently). And removed the pkg/ mention as the path do not exist anymore.

PR #787, opened 2026-07-27
Container

vassilit:
rmlint doesn't delete files by itself; it generates scripts and lists. (but --dedupe can modify the filesystem)
But for testing, and for anyone who'd still prefer an extra layer of safety or don't want to bother with dependancies, it can now also be run in a container with bind mounts.


RayOei (commented):
I think it would be helpful to add some information to the `development.rst` about running and using the container? A quick example of how to use the bind mount?
I quickly checked with podman (`podman build -t tagname .`  and `podman run tagname:latest`) and docker (`docker build -t tagname -f Containerfile .` and `docker run -tagname: latest`)

RayOei (approved):
👍

PR #798, opened 2026-08-26
formats/py: fixes

### LINKED ISSUE #797: py format paranoid check does not work
The py format function original_check does not pass `shallow=False` to `filecmp.cmp`:
https://github.com/sahib/rmlint/blob/b311f38dab01585a138ed0f3ac14d2b426d12523/lib/formats/py.py#L72

Per https://docs.python.org/3/library/filecmp.html, the default behavior of filecmp.cmp checks only the file type, size, and modification time, not the full contents of the file.

Steps to reproduce:
```
/tmp$ mkdir rmlint-bug
/tmp$ cd rmlint-bug/
/tmp/rmlint-bug$ echo "foo" > file1
/tmp/rmlint-bug$ cp -a file1 file2
/tmp/rmlint-bug$ rmlint --types=duplicates -o py:rmlint.py .
/tmp/rmlint-bug$ echo "bar" > file2
/tmp/rmlint-bug$ touch -r file1 file2
/tmp/rmlint-bug$ ./rmlint.py -d -p
[100%] Keeping original:   /tmp/rmlint-bug/file1
[100%] Deleting duplicate: /tmp/rmlint-bug/file2
[100%] Done!
```

This was reproduced with the following version, although the issue appears to still exist in the current source code.
```
version 2.10.2 compiled: Mar 12 2025 at [17:09:22] "Ludicrous Lemur" (rev 0.2)
compiled with: +mounts +nonstripped +fiemap +sha512 +bigfiles +intl +replay +xattr +btrfs-support
```

vassilit:
filecmp.cmp() defaults to shallow=True, which returns True without reading a single byte if os.stat() signatures match.

Closes: #797

### DIFF
--- a/lib/formats/py.py
+++ b/lib/formats/py.py
@@ -33,13 +33,12 @@ import argparse
 import filecmp
 import json
 import os
-import pwd
 import shutil
 import subprocess
 import sys
 
 CURRENT_UID = os.geteuid()
-CURRENT_GID = pwd.getpwuid(CURRENT_UID).pw_gid
+CURRENT_GID = os.getegid()
 
 USE_COLOR = sys.stdout.isatty() and sys.stderr.isatty()
 COLORS = {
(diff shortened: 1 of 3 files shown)


vassilit:
@RayOei this one is ready as well

RayOei (approved):
👍

PR #803, opened 2026-09-02
Fix failing test using localtime

RayOei:
Fixes failing test as gmtime doesn't work when you are not on GM time 😁

### DIFF
--- a/tests/test_options/test_newer_than.py
+++ b/tests/test_options/test_newer_than.py
@@ -41,7 +41,7 @@ def test_simple(usual_setup_usual_teardown):
     for offset, expect in [(1, 3), (8, 0)]:
         iso_time = time.strftime(
             "%Y-%m-%dT%H:%M:%S%z",
-            time.gmtime(now + offset)
+            time.localtime(now + offset)
         )
 
         head, *data, footer = run_rmlint_once('-S a -N ' + iso_time)


vassilit:
Something is fishy nevertheless as the test pass everywhere else, because:

```
tests/test_options/test_newer_than.py::test_simple
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/vt/dev/rmlint/main/tests/test_options/test_newer_than.py(44)test_simple()
-> breakpoint()
(Pdb) iso_time
'2026-09-02T15:17:41+0000'
```

We pass timezone-aware time (+0000) and in rmlint, we use Glib g_date_time_new_from_iso8601() which parse timezone.

So _gmtime_ or _localtime_ shouldn't make any difference and that is what we see in the tests.

vassilit:
> Fixes failing test as gmtime doesn't work when you are not on GM time 😁

Could you document why in the commit message ?

RayOei:
This essentially shows the issue: I logged the files created and the `iso_time` created in `test_simple` loop.
The _actual_ files created are also listed, with the, for me, correct time. The `iso_time` is an hour off with the TZ.  

> ------------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------------
> /private/var/folders/gc/b82fxvzj0pdfzcnz_1cf_t6c0000gn/T/pytest-of-ray/pytest-16/test_simple0/a
> /private/var/folders/gc/b82fxvzj0pdfzcnz_1cf_t6c0000gn/T/pytest-of-ray/pytest-16/test_simple0/b
> /private/var/folders/gc/b82fxvzj0pdfzcnz_1cf_t6c0000gn/T/pytest-of-ray/pytest-16/test_simple0/c
> 1: 2026-09-02T14:50:40+0100
> 2: 2026-09-02T14:50:57+0100
> 
> 
> (pyenv) ....:/private/var/folders/gc/b82fxvzj0pdfzcnz_1cf_t6c0000gn/T/pytest-of-ray/pytest-current/test_simple0 » ll
> total 20
> drwx------ 7 ray  224 Sep  2 16:50 .
> drwx------ 4 ray  128 Sep  2 16:50 ..
> -rw-r--r-- 1 ray  228 Sep  2 16:50 .pretty-0
> -rw-r--r-- 1 ray    3 Sep  2 16:50 a
> -rw-r--r-- 1 ray    3 Sep  2 16:50 b
> -rw-r--r-- 1 ray    3 Sep  2 16:50 c
> -rw-r--r-- 1 ray 1956 Sep  2 16:50 out.json

RayOei:
I don't particularly likes these combined tests as it makes it harder to see what the actual intent and/or failure is.
I see that with more tests which I normally would split in single checks with also more verbose descriptions. 

The test_simple has actually two checks:
- whether a very small time difference from creation, almost equal to `now`, will return the three created files
- whether a larger time difference from `now` will return 0 files as their created time is before the filtered time.
The last one fails with the gmtime as it defines a filter an hour _before_ `now`, so the three files are returned and the expect(ed) = 0. So it fails.

This check is therefore dependant on the interpretation of gmtime and the actual localtime the files are created in, which apparently works for CI. But not for me locally. The current TZ difference between GMT and CEST is 2 hours.

vassilit (approved):

vassilit:
I agree this test the way is written is a bit precarious.

I was afraid this might be a rmlint bug on macOS, not a test to fix, so I wanted to understand *why* because it wasn't clear to me.

In particular: 
> 1: 2026-09-02T14:50:40+0100
> 2: 2026-09-02T14:50:57+0100

What are those ?

Turns out there are UTC timestamp with a DST marker applied to them. I was not aware that gmtime can be interpreted that differently across platforms.

J'ai renversé 1L d'eau minérale sur mon macbook, je l'ai donc complètement desossé et je n'ai pas d'autre Mac ici à ma disposition. Ceci dit, je devrais plutôt te parler breton et non français ;)
But I think I found a good explaination to the bottom of this without needing to have the machine itself:
https://gist.github.com/macintux/78516cba33cf8d58868917b7043d2d50

Apple and FreeBSD, which share code for the libc,

https://github.com/apple-open-source-mirror/Libc/blob/master/stdtime/FreeBSD/strftime.c#L517
https://github.com/freebsd/freebsd-src/blob/main/lib/libc/stdtime/strftime.c#L495

but Apple, wanting to be UNIX compliant, added a __DARWIN_UNIX03 macro (defined by default) that force the C99-buggy route. But FreeBSD rely solely on TM_GMTOFF, which is defined by default as well.
GNU has a completly different implementation (_GNU's Not UNIX_).

The CI was not impacted because its localtime IS UTC, so `tm_isdst` is 0.

One alternative to better test rmlint (Glib in fine) parsing of time would have been to force a non-UTC ISO time with DST:

```python
from datetime import datetime, timedelta, timezone
from zoneinfo import ZoneInfo

BXL = ZoneInfo("Europe/Brussels")

[...]

datetime.fromtimestamp(now + offset, tz=timezone.utc).astimezone(BXL).strftime("%Y-%m-%dT%H:%M:%S%z")
```

RayOei:
Nice research 😁 

Ouch... a drowned mac

Je comprends plus ou moins le français. Le breton, c'est complètement différent…

PR #804, opened 2026-09-03
Update readme

RayOei:
Update authors - I think the amount of commits warrants adding vassilit to the list 👍 
Format fix
Advertise activity on `develop` and `master` - which will show as soon as this is merge _to_ master.


RayOei:
Oops...

PR #805, opened 2026-09-03
Update readme

RayOei:
Update authors - I think the amount of commits warrants adding vassilit to the list 👍
Format fix
Advertise activity on develop and master - which will show as soon as this is merged to master.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -5,3 +5,12 @@ Christopher Pahl <sahib@online.de> <chris@modi.localdomain>
 Christopher Pahl <sahib@online.de> <b@mmel.localdomain>
 Christopher Pahl <sahib@online.de> <sahib@localhost.localdomain>
 Christopher Pahl <sahib@online.de> <chris@vbox.(none)>
+Christopher Pahl <sahib@online.de> <wohngemeinschaft@localhost.local>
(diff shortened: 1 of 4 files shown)


vassilit:
Thank you for adding me. On one hand it awaken my imposter syndrome, on the other hand it is probably useful to mark the project as not-totally-abandoned. In that regard, I've started in reality working on rmlint in late 2024 (like, the very end of December, e.g. 3rd of January 2025 in #672), but was not pushing anything back then, I was more testing things locally; and I know for a fact that it's real work as well. At this stage, you could very well add yourself as a core tester and/or contributor.
Also, I'd prefer using my full name rather than a Github handle (keep it in URL, of course!).

I don't think also we should add a badge about master: it is not an active branch per se but rather a regular snapshot of develop when things are deemed stable. The badge visible on master should reflect develop activity as well IMHO.

RayOei:
The in the readme mentioned AUTHORS list doesn't seem to exist, as far as I can make out.
For me I feel it is too early to be in the main authors list, maybe when I get in the +50 commit range or so 😀 
You are right: as soon as master is up-to-date with develop the information on the landing page should be fine.

vassilit (approved):

PR #809, opened 2026-09-04
2.11.0-dev

vassilit:


RayOei (approved):
Looks good but I have still/again issues with running GUI on my mac. So I can't sanity check that at this point.
I remember you mentioned a PR you may already have prepared?

vassilit:
> Looks good but I have still/again issues with running GUI on my mac. So I can't sanity check that at this point. I remember you mentioned a PR you may already have prepared?

More or less, you'd have to reproduce what the CI does:
```
brew install scons pkgconf glib json-glib
pip install setuptools build installer
PYTHON=python3 scons install
```
But there are a lot of variables to consider: did you install Python using Homebrew or directly from python.org? Do you need root privileges to install packages into site-packages? What does your PYTHONPATH look like?

On macOS, there are so many possible configurations that things quickly become complicated, and every installation can be slightly different.

It may seem counterintuitive, but Linux and BSD systems are generally much simpler and more homogeneous in this regard.

vassilit:
We produce a standard Python wheel. That can be installed with any tool (pip, uv, etc). We use build and installer.

It seems that for mac, maybe a little more is [needed](https://gnome.pages.gitlab.gnome.org/pygobject/guide/deploy.html).

vassilit:
Would you be interested in working on this ?
Do you have a paid Apple Developer account ?

vassilit:
> Looks good but I have still/again issues with running GUI on my mac. So I can't sanity check that at this point. I remember you mentioned a PR you may already have prepared?

1. Do you have issue with GUI with the version actually on master (2.10.3) ? I.e. develop works _worse_ than master ?
2. For this version (develop), I think the better way to install is something like `PYTHON="$(brew --prefix)/bin/python3" scons install` from a fresh checkout and with dependancies already installed. This is if you installed Python with homebrew. If not, `PYTHON=/Library/.../python3 scons install` (replace by your path)

I'll repair my mac, so that I can test (I could try a VM, but I would like the laptop working anyway). The Macbook is running Linux, but I have a small macOS Monterey partition. It's an Intel mac, so what works for me might not for you (e.g. `/usr/local` vs `/opt/homebrew`)

Unfortunatly, neither of the ports install the GUI:

[Homebrew](https://formulae.brew.sh/formula/rmlint)

[Macports](https://ports.macports.org/port/rmlint/details/)

PR #812, opened 2026-09-07
gui: refresh algorithms list to have rmlint actually supports

vassilit:
We recently changed the schema ID, let's use that opportunity to tidy up.


RayOei (approved):
I can't actually sanity check at this time.
But I like the scheme 👍

PR #818, opened 2026-09-09
docs: fix badges

vassilit:
A broken badge image link made EPUB generation fails. Fix it and don't include badges into formats other than HTML.

Test EPUB in CI because RTFD doesn't do it for PRs (HTML only).

Put the badges into a separate file, so that packagers can easily remove it for privacy patches, like https://salsa.debian.org/debian/rmlint/-/blob/debian/master/debian/patches/privacy-breach-workaround.patch?ref_type=heads

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -42,6 +42,7 @@ jobs:
       - name: Build documentation
         run: |
           scons STRICT=1 docs
+          sphinx-build -T -W -b epub -d docs/_build/doctrees docs docs/_build/epub
 
       - name: Prepare test environment
         # The test suite is seriously disk-intensive. Given that linux
--- a/README.rst
+++ b/README.rst
@@ -10,20 +10,36 @@ offers to remove it.
 .. image:: https://img.shields.io/github/v/release/sahib/rmlint?include_prereleases&display_name=release
    :target: https://github.com/sahib/rmlint/releases
 
-.. image:: https://github.com/sahib/rmlint/actions/workflows/build-and-test.yml/badge.svg
-   :target: https://github.com/sahib/rmlint/actions
-
-.. image:: https://readthedocs.org/projects/rmlint/badge/?version=latest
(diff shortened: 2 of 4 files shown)


RayOei:
Am I missing the `_badges.rst`?

vassilit:
Yes I forgot to git add :) as I was in a hurry to catch my transport as I had to go somewhere.
I will do it when back home.

RayOei:
No problem 😁

RayOei:
Ha.. 👍 for the factorizing... I was looking into the same 😁 
There is one other place the authors need updating: rmlint.1.rst

Also the translations and lib/cmdline.c are behind, I already updated those but I wait for this one to incorporate the latest changes.

vassilit:
Yeah, but no. See the [README](https://github.com/sahib/rmlint/tree/a576f2784011437c507615b62de2a52aeafe8f8c). Github does [not](https://github.com/github/markup/issues/172) support includes.

Also, I've tried to include _licence.rst in the manpage, but the links are not pretty there.

RayOei:
Any idea why RTD compile fails for `master`?

RayOei:
Or is the missing `https://github.com/sahib/rmlint/actions/workflows/build-and-test.yml/badge.svg` the reason? So wait for this merge?

RayOei:
NVM... 

>4 warnings (with warnings treated as errors).

vassilit:
This is the best that I can do.

And yes, fail_on_warning is true (default: false) in our .readthedocs.yaml because I think it's better than silently failling.
Maybe it was a bad idea. What do you think ?

vassilit:
> This is the best that I can do.

Means that I won't push anymore. Either keep it that way or try something else :)

master will be fixed in a few days with this PR in the release candidate. I think to add set -e to the shell script and call it a release already.

RayOei:
> Maybe it was a bad idea. What do you think ?

No, it is fine. Now I am aware of that 😁

RayOei (approved):
👍

PR #821, opened 2026-09-09
i18n: honour TEXTDOMAINDIR on debug

vassilit:
Test a translation:

```bash
$ mkdir -p locale/ka/LC_MESSAGES
$ msgfmt po/ka.po -o locale/ka/LC_MESSAGES/rmlint.mo
$ TEXTDOMAINDIR=locale LANG=ka_GE.UTF-8 LC_MESSAGE=ka_GE.UTF-8 ./rmlint --version
version 2.11.0-dev+g0491702a "Evolving Echidna"
built Sep  9 2026 at 16:12:17 UTC
აგებულია რით: +mounts +nonstripped +fiemap +intl +xattr +btrfs-support
[...]
```

You still need:
```bash
# sed -i '/ka_GE.UTF-8/s/^# //' /etc/locale.gen
# locale-gen
```
(or on Debian: `dpkg-reconfigure --frontend=noninteractive locales`)

### DIFF
--- a/src/rmlint.c
+++ b/src/rmlint.c
@@ -62,7 +62,13 @@ static void signal_handler(int signum) {
 static void i18n_init(void) {
 #if HAVE_LIBINTL
     /* Tell gettext where to search for .mo files */
+#ifdef RM_DEBUG
+    const char *locale_dir = getenv("TEXTDOMAINDIR");
+    bindtextdomain(RM_GETTEXT_PACKAGE,
+                   locale_dir ? locale_dir : INSTALL_PREFIX "/share/locale");
+#else
     bindtextdomain(RM_GETTEXT_PACKAGE, INSTALL_PREFIX "/share/locale");
+#endif
     bind_textdomain_codeset(RM_GETTEXT_PACKAGE, "UTF-8");
 
     /* Make printing umlauts work */


RayOei:
> ```shell
> mkdir -p locale/ka/LC_MESSAGES
> ```

Is this from the development directory?

vassilit:
> > ```shell
> > mkdir -p locale/ka/LC_MESSAGES
> > ```
> 
> Is this from the development directory?

yes. Any directory. rmlint.mo must be in locale/ka/LC_MESSAGES and locale by specified in TEXTDOMAINDIR=

RayOei:
Hmmm it doesn't work for me 😕 
I also see the default `/share/locale` but that doesn't exist on macOS.

vassilit:
```bash
$ scons DEBUG=1
$ export LANGUAGE=de
$ mkdir -p locale/$LANGUAGE/LC_MESSAGES
$ msgfmt po/$LANGUAGE.po -o locale/$LANGUAGE/LC_MESSAGES/rmlint.mo
$ TEXTDOMAINDIR=locale ./rmlint --version
kompiliert mit:
```

If that does not work, do `export LC_MESSAGES=en_US.UTF-8` or replace en_US.UTF-8 with whatever locale is installed into your system libc (that's bit hacky, but should work).

RayOei:
Hmmmm interesting...
Running the tests fails as `test_fdups_and_traversed_dirs_in_summary` tests for `1 other suspicious item(s)` but it returns NL now 😁 
Running `rmlint` on commandline still uses `en` and not `nl` while I switched the system language to Dutch/NL too.
Checked for duplucate applications, that is also not the case. I am a bit stumped... 😕

vassilit:
> Hmmmm interesting... Running the tests fails as `test_fdups_and_traversed_dirs_in_summary` tests for `1 other suspicious item(s)` 

Probably the tests needs to run with `C.UTF-8` always and we should create a couple of tests that sets the locale to something else and test some translations.

RayOei:
> test some translations

Yes for sure. 
But I am a bit mystified why the test seems to get translated results while running afterwards doesn't. I will experiment a bit further. I have a separate Dutch language branch, not pushed yet.

RayOei:
Even having set the `LC_MESSAGES` the `locale` is still `en_US-UTF-8` (arrow is me).

```
@../Projects/rmlint(test_translations⚡) » printenv
...
VSCODE_PROFILE_INITIALIZED=1
VSCODE_PYTHON_AUTOACTIVATE_GUARD=1
-> LC_MESSAGES=nl_NL.UTF-8

@../Projects/rmlint(test_translations⚡) » locale
LANG="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_CTYPE="en_US.UTF-8"
-> LC_MESSAGES="en_US.UTF-8"*
LC_MONETARY="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_ALL="en_US.UTF-8"
```

The only way to get it to work seems to use `export LC_ALL=nl_NL.UTF-8` as by default it is always US, even when I have set my macOS language to NL 😕 
This may be a macOS quirk... digging deeper...

vassilit:
I must go on a trip soon, so I will definitely reassemble my Mac, which must have dried out by now.
I'll try there.

RayOei:
With the LC_ALL I at least get a translation BUT it always takes it from `/usr/local/share/locale`.
So setting the `TEXTDOMAINDIR` doesn't make a difference.
I also noticed that there are many translations missing, not defined to be translated in the code. I will go over those and update the `.pot`

And safe travels! I am on my way home, working in the train. I had too little time in Paris, so that drink is for another time!

RayOei:
BTW: `TEXTDOMAINDIR=locale ./rmlint -h` does work in my Ubuntu VM.

RayOei:
GUI has only parts translated, using 

`
TEXTDOMAINDIR=locale PATH="$PWD:$PATH" GSETTINGS_SCHEMA_DIR="$PWD/gui/schemas" shredder 
`

Scherm­afbeelding 2026-09-18 om 21 43 59

PR #822, opened 2026-09-09
hash-utility: fix stdin

vassilit:
Terminate the array.
Warn on invalid path.

Closes: #402

PR #824, opened 2026-09-10
Make install nicer for packagers

vassilit:


vassilit:
After this one and #825, #822 (we should write a test someday, but the fix is OK so we can merge), and make sure CHANGELOG.md is somehow up-to-date (it might not be exhaustive), we can release 2.11.0-rc.0

RayOei (approved):
👍

PR #825, opened 2026-09-11
treemerge: fix honouring directories layout

### LINKED ISSUE #742: honour-dir-layout doesn't work
folder structure:
```
── aa/
   ├── bb/
   │   └── yy.txt
   └── xx.txt
── cc/
   └── bb/
       ├── xx.txt
       └── yy.txt
```
With `rmlint aa cc -D`, it removes the entire "cc" folder. OK.

With `rmlint aa cc -Dj`, same result.
But the -j option should recognize a different tree structure, and mark no folder as duplicate. 
The -j option doesn't work, or I am missing something?

vassilit:
It didn't record the association between filenames and their content, so a directory with swapped content but identical filenames or just identical filenames but at differents places was reported as a duplicate.

Closes: #742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format follows [keepachangelog.com]. Please stick to it.
 * FreeBSD: ``rmlint.sh``'s RMLINT_BINARY= entry.
 * Fix ``--honour-dir-layout``/``-j`` when filenames and content are the sames, just not
   at the same place.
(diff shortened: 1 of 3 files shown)


RayOei (approved):
👍 
As a minor note: I prefer tests per `rmlint` call as to distinguish a bit better between different outcomes, unless sequential calls are specifically needed for that test, obviously.

PR #827, opened 2026-09-12
rmlint.sh: abort on errors by default

### LINKED ISSUE #469: Add set -e to the bash script?
In general, set -e is not such a great thing, as it requires extra care when working with pipes.
(http://mywiki.wooledge.org/BashFAQ/105)

However, rmlint.sh does not seem to be using a lot of pipes, and generally is using a very fragile technique of bash script generation, so I expect set -e to bring more good than harm, in case some strange file names are escaped incorrectly, for example.

vassilit:
Closes: #469 
(the "closes" might not work because develop is not the default branch. I don't know if it works once indirectly merged to master)


RayOei (approved):
👍

PR #829, opened 2026-09-14
fix: do not --backup-rename special output paths

### LINKED ISSUE #763: --output=stamp:/dev/null deletes /dev/null
In trying to get rmlint to output nothing (https://github.com/sahib/rmlint/issues/762), I did a run with `--output=stamp:/dev/null` to try to have it write its output to /dev/null.

To my surprise, rmlint deleted /dev/null then created a plain-text file in its place.

I recovered my system, thankfully, but this seems like quite a surprising footgun.

vulragrag-star:
## Summary

`--backup` no longer renames special output paths such as `/dev/null`.

Fixes #763.

## Why

When `--backup` was set and an output path already existed, rmlint renamed the old path aside before opening it for write. For `/dev/null` that moved the device node and left a regular file in its place.

## Change

Only regular files are renamed for backup. Devices, symlinks, and other non-regular paths are left alone; `fopen` still writes to them as before.

## Tests

```console
$ python3 -m pytest tests/test_formatters/test_backup_special.py -q
```

Both cases passed (symlink-to-`/dev/null` and `/dev/null` itself stay intact under `--backup`). A manual check also confirmed regular stamp files still get a timestamped backup.


lib/formats.c
```diff
@@ -27,6 +27,8 @@
 
 #include <ctype.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
```
vassilit: Everything else is good, but why `unistd.h` ?

vulragrag-star: Good catch — it isn't needed.

`lstat` / `S_ISREG` come from `sys/stat.h`. Dropped the unused `unistd.h` in 4d39dac0.

vassilit (approved):

PR #830, opened 2026-09-18
Prerelease 2.11.0-rc.0

vassilit:
Same thing, `--ff-only`

(I could tag on develop, but that's not pretty)


RayOei (approved):
👍 

There we go 😀

PR #831, opened 2026-09-19
Android experimental support

vassilit:
Based on the work done in https://github.com/termux/termux-packages/pull/31682


DevGitPit:
- **Default prefix on Termux:** When running `scons` without arguments on Termux, SCons's `PathVariable.PathIsDir` check currently fails because the default `/usr/local` does not exist on Android. In `SConstruct`, `get_default_prefix()` could check for Termux:
  ```python
  if 'TERMUX_VERSION' in os.environ:
      return os.environ.get('PREFIX', '/data/data/com.termux/files/usr')
  ```
  Gating it on `TERMUX_VERSION` (or `not Path(DEFAULT_PREFIX).is_dir()`) prevents any unintended side effects on standard Linux setups while allowing `scons` to work out-of-the-box in Termux without requiring `PREFIX=$PREFIX`.

- **Shredder temp directory:** In `gui/shredder/logger.py`, the logger currently hardcodes `/tmp/app.log`. Because `/tmp` does not exist on Android, using `os.path.join(tempfile.gettempdir(), 'shredder.log')` would allow Shredder to run out-of-the-box on Android (where Python resolves `tempfile.gettempdir()` via `$TMPDIR`).

vassilit:
> * **Default prefix on Termux:** `TERMUX_VERSION` [...] `PREFIX=$PREFIX`.

Well, require that the PREFIX (whatever it comes from) for targets that are not related to installing is wrong anyway, We will probably remove SCons's default validator `PathVariable.PathIsDir` and validate the path on install targets only.

If looking in os.environ, $TERMUX_PREFIX I naively understand is used for cross-compilation from Ubuntu and $PREFIX in Termux ?

> * **Shredder temp directory:**

Was fixed in 020948c. Probably a good idea to lower the default verbosity as well.

--- 
I could run the full testsuite in CI but that required compiling the blake3 Python wheel with Rust and emulating Python `os.link` using Bionic directly.

DevGitPit:
$PREFIX is set by the Termux runtime in shell sessions; $TERMUX_PREFIX is only set inside termux-packages build scripts (cross and on-device build-package.sh).

DevGitPit:
> Well, require that the PREFIX (whatever it comes from) for targets that are not related to installing is wrong anyway, We will probably remove SCons's default validator PathVariable.PathIsDir and validate the path on install targets only.

Unless it is later baked in the binary or any generated files. But, yes removing this validation during build is way better an option.

vassilit:
> > Well, require that the PREFIX (whatever it comes from) for targets that are not related to installing is wrong anyway, We will probably remove SCons's default validator PathVariable.PathIsDir and validate the path on install targets only.
> 
> Unless it is later baked in the binary or any generated files. But, yes removing this validation during build is way better an option.

It is actually baked for gettext, that Termux does not support (AFAIK), so. Also, we could use relative paths like we do now in the GUI.

